Issue #73, opened 2023-08-23
Bug: v1.5.0 - Typebot messages messages aren't displayed in chatwoot


franciscoeymardneto:
The visibility of Typebot messages within the Chatwoot platform. Currently, messages sent by the Typebot are not being displayed within Chatwoot, creating a disconnect in the conversation history.

![Captura de Tela 2023-08-23 às 10 03 26](https://github.com/EvolutionAPI/evolution-api/assets/78768690/82298218-4b60-4032-a0eb-398a9be9d3a7)

DavidsonGomes:
corrigido na versao 1.5.1

PR #74, opened 2023-08-23
fix: Typebot Self Activation and Chatwoot Message Visibility Issues

franciscoeymardneto:
This pull request aims to address two critical issues concerning the interaction of the Typebot with our integrated WhatsApp messaging system, specifically within the Chatwoot platform.

Issues related: #71 #73

### Typebot starting itself 

- [X] Identify the cause of the issue
- [X] prevent typebot to consider messages from the sender 
- [ ] tests

### Typebot messages within chatwoot

- [X] Identify the cause of the issue
- [ ] send the typebot messages also to chatwoot 
- [ ] tests


DavidsonGomes:
This has already been addressed in develop thanks!

PR #78, opened 2023-08-26
clean: docker network declaration and build of the `evolution/api` image

AdsonCicilioti:
SH code to check existence of unnecessary docker network.

 ```yaml
    external: true
 ```
 It is only for using an existing main network.
 Since the `docker-compose*.yml` files are already primarily responsible for provisioning the main application stack, it doesn't make sense to create the network in a task outside its scope.

 This is just a small contribution from someone who has just arrived and is already impressed with this community!

The docker build command will be removed in future releases in favor of [buildx](https://docs.docker.com/build/architecture/#buildx).
 The Dockerfile build can be done by `docker-compose*.yml` itself by adding the `build: .` property in the api service.

### DIFF
--- a/Docker/mongodb/docker-compose.yaml
+++ b/Docker/mongodb/docker-compose.yaml
@@ -35,7 +35,8 @@ volumes:
   evolution_mongodb_data:
   evolution_mongodb_configdb:
 
+
 networks:
   evolution-net:
-    external: true
-    
+    name: evolution-net
+    driver: bridge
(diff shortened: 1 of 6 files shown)


DavidsonGomes (approved):

AdsonCicilioti:
@DavidsonGomes if the instruction to use the `./docker.sh` file exists somewhere in the documentation, you will have to update to just use `docker compose up -d`.

PR #105, opened 2023-09-09
Env

gabrielpastori1:
1. New Environment Variables: Introduced two new environment variables to augment our application's flexibility:
-  `HIDE_INDEX`: A feature to hide the index, with a default value of false.
- `HIDE_MANAGER`: A feature to hide the manager interface, also defaulting to false.

2. CORS Default Value Adjustments: Rectified the default values for `CORS_ORIGIN `and `CORS_METHODS `within the envProcess.

3. Port Configuration Enhancement: `SERVER_PORT `or the more commonly used `PORT `can be employed to specify the server's listening port.  To foster broader compatibility, especially with platforms like Heroku.

PR #107, opened 2023-09-10
Heroku button

gabrielpastori1:


DavidsonGomes:
Very good, explain more about this improvement and also make sure that the other open pull #105 request is within this one

gabrielpastori1:
@DavidsonGomes

This update prepares the project to be easily sent to Heroku, some configurations in the `env.config.ts` file to allow this easier deployment and maintain the same behavior in other cases. The `app.json` file was also created which contains project and env specifications, a `Procfile` file was also added pointing out which script Heroku should use to start the dynos. 

In the readme.MD file I placed an installation section with the heroku button and the VPS recommendation that I found in the documentation. 

The other PR #105 is already inside this one so I'm going to delete it, in it I added 2 new ENV, `HIDE_INDEX` and `HIDE_MANAGER` both by default are false and disable the route `/` and `/manager`

PR #108, opened 2023-09-11
[Improvement] - Send and Update source_id of messages to chatwoot

raimartinsb:
Very well tested and for security I created a new environment variable so that there is no impact from this new improvement, which will Send and Update source_id of messages to chatwoot


DavidsonGomes:
Great, before accepting, explain to me exactly about this improvement

raimartinsb:
> Great, before accepting, explain to me exactly about this improvement

In my work we have a "custom chatwoot", with reply capabilities, for this we need the source_id of the messages, with my PR this can fill this info on chatwoot. If the user of Evolution API does nothing, nothing was happen. But, if the environtment USE_REPLY_ID is seted to true, on replyes the message set the source_id on chatwoot.

DavidsonGomes:
Merge reversed, problems with methods and fields not existing in the lib @figuro/chatwoot-sdk

PR #198, opened 2023-10-29
Handle erros in Typebot

gabrielpastori1:
Refactor TypebotService to handle error responses from axios

The TypebotService in `typebot.service.ts` file has been refactored to handle error responses from the axios library. Error handling has been added to the `startTypebot`, `createNewSession`, and other related methods to catch and log errors from failed axios requests.


sebavieira:
Realmente isso ajudou de mais, obrigado pela correção.

PR #214, opened 2023-11-12
message.type === "image" in typebot.service.ts

suissa:
Only works, for me, in this way.

Added ".png" in mediaMessage.media

If it`s only for me disregard.


DavidsonGomes:
As we have several usage methods, we cannot implement something so specific

PR #331, opened 2023-12-28
refactor(chatwoot): remove message ids cache in chatwoot to use chatwoot's api itself

jaison-x:
refactor(chatwoot): remove message ids cache in chatwoot to use chatwoot's api itself

Remove use of disc cache to optimize performance.

BREAKING CHANGE: to make this, we need to use the param `source_id` from message in chatwoot.
This param is only available from api in chatwoot version => 3.4.0.


jaison-x:
> my cpu thanks you, great job works fine so far the change you are referring to is this, https://github.com/chatwoot/chatwoot/pull/8359/files for those who want to try in versions <=3.4.0

Good Nestor, with this everybody with version <= 3.4.0 can use this PR too! This PR helps a lot with performance in chatwoot integration.

PR #370, opened 2024-01-18
fix: Sending status message

leandrosroc:
Fixed when requested to publish status message, it goes into the condition of not being audio and also the jid is defined as "+status@broadcast"


gabrielpastori1 (approved):

lindenbergp:
Já está no docker? Queria testar. @DavidsonGomes .

PR #371, opened 2024-01-20
Add number parameter to OnWhatsAppDto constructor

judsonjuniorr:
When requesting if the Whatsapp number is valid, if the number sent contains the 9 digit from Brazil, and the result comes without that digit, there's no way to identify the number sent inside the response array.

With this change the response DTO will become like this:
image

With that when the API returns the list of valid numbers I can identify that exact number I sent

### DIFF
--- a/src/whatsapp/dto/chat.dto.ts
+++ b/src/whatsapp/dto/chat.dto.ts
@@ -1,7 +1,12 @@
 import { proto, WAPresence, WAPrivacyOnlineValue, WAPrivacyValue, WAReadReceiptsValue } from '@whiskeysockets/baileys';
 
 export class OnWhatsAppDto {
-  constructor(public readonly jid: string, public readonly exists: boolean, public readonly name?: string) {}
+  constructor(
+    public readonly jid: string,
+    public readonly exists: boolean,
+    public readonly number: string,
+    public readonly name?: string,
+  ) {}
 }
 
 export class getBase64FromMediaMessageDto {
--- a/src/whatsapp/services/whatsapp.service.ts
+++ b/src/whatsapp/services/whatsapp.service.ts
@@ -3137,9 +3137,9 @@ export class WAStartupService {
 
         if (!group) throw new BadRequestException('Group not found');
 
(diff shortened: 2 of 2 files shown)


leandrosroc:
With the exact number returned of what was sent, will it be used for comparison or for later sending? because if the jid returned without the ninth digit, then it would later be sent to the person who is in the jid, being registered on WhatsApp

leandrosroc:
Jid Whatsapp is the most correct number to use when sending

judsonjuniorr:
@leandrosroc will be used just for comparison.

Here an example:

Request:
```json
{
  "numbers": [
    "5534998313925",
    "553498313925"
  ]
}
```

Response:
```json
[
    {
        "jid": "553498313925@s.whatsapp.net",
        "exists": true,
        "number": "5534998313925"
    },
    {
        "jid": "553498313925@s.whatsapp.net",
        "exists": true,
        "number": "553498313925"
    }
]
```

And then I can compare to save the jid in the database for that number:
```js
[
    "5534998313925",
    "553498313925"
].map(number => {
    const jidValidation = result.find(r => r.number === number)
    const jidExists = jidValidation?.exists ?? false
    const jid = jidValidation?.jid
})
```

And with this I can store the jid on the database for that specific number

leandrosroc:
I understood.
What if you used the .map function like this?

```
[
    "5534998313925",
    "553498313925"
].map(number => {
    const jidValidation = result.find(r => r.jid=== `${number}@s.whatsapp.net`);
    const jidExists = jidValidation?.exists ?? false
    const jid = jidValidation?.jid
})
```

leandrosroc:
I understood

judsonjuniorr:
I tried but It's not the exact value, check this example:

Request:
```json
{
  "numbers": [
    "5534998313925",
  ]
}
```
Response:
```json
[
    {
        "jid": "553498313925@s.whatsapp.net",
        "exists": true,
    }
]
```

You can see I sent `5534998313925` but the jid number returned is `553498313925` without the 9 digit

leandrosroc (approved):

leandrosroc:
I agree, makes sense for the proposed objective

PR #390, opened 2024-01-31
feat: If contact stored with name, name added in return

leandrosroc:
In `chat/whatsappNumbers` and `chat/fetchProfile` routes, name will be filled if stored
197d1a29-bb1a-4e80-b216-8488d7fb7fad
67aafc36-113a-426a-923f-e5f694005686

### DIFF
--- a/src/whatsapp/services/whatsapp.service.ts
+++ b/src/whatsapp/services/whatsapp.service.ts
@@ -3130,7 +3130,7 @@ export class WAStartupService {
     const jids: {
       groups: { number: string; jid: string }[];
       broadcast: { number: string; jid: string }[];
-      users: { number: string; jid: string; contact?: string; name?: string }[];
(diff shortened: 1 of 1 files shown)


judsonjuniorr (changes requested):
It has been added a new property to the users array `contact` that only store the jid, why not only use the `user.jid` instead of creating a new one with the same value?

You could also make the `repository.contact.find` outside the map and find all contacts first in one single query, after that on the map loop you just do a `.find` in the result

leandrosroc:
the jid in users starts with "+"

judsonjuniorr:
> the jid in users starts with "+"

You can remove it for example: `"+55 11 9 9999-9999".replace(/\D/g, '')` with this you'll get only the number values from the string

leandrosroc:
using "replace", it also removes @s.whatsapp.net

PR #479, opened 2024-03-19
feat(endpoint): add setPresence endpoint

dv336699:
This endpoint will allow you to set an instance presence without a phone number.

image

### DIFF
--- a/src/validate/validate.schema.ts
+++ b/src/validate/validate.schema.ts
@@ -161,6 +161,18 @@ export const presenceSchema: JSONSchema7 = {
   required: ['options', 'number'],
 };
 
+export const presenceOnlySchema: JSONSchema7 = {
+  $id: v4(),
+  type: 'object',
+  properties: {
+    presence: {
+      type: 'string',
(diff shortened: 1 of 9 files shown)


DavidsonGomes:
Sorry, but what is the application? and how does it differ from SendPresence?

Screenshot 2024-03-19 at 10 07 43

dv336699:
Hi,

Thanks for considering this. 

The method from Baileys, `SendPresence`, doesn't require a phone number.

The endpoint `chat/SendPresence` enforces a phone number.

As per https://github.com/WhiskeySockets/Baileys?tab=readme-ov-file#update-presence you can call `SendPresence` without a number to mark the device as unavailable.

> Note: In the multi-device version of WhatsApp -- if a desktop client is active, WA doesn't send push notifications to the device. If you would like to receive said notifications -- mark your Baileys client offline using sock.sendPresenceUpdate('unavailable')

Here is a snippet from the WhatsApp group with more details.

image


I didn't want to touch the `chat/sendPresence` endpoint as to avoid breaking changes in the future.

Let me know if you have any more questions.

Thanks

PR #577, opened 2024-05-06
feat: method to mark chat as unread

neanderdev:
I added a method to mark the chat as unread.


bergpinheiro:
Como usa essa função? Qual endpoint?

neanderdev:
O endpoint /chat/markChatUnread/instanceName

PR #739, opened 2024-08-06
Improve reconnection logic for ban and block cases and add disconnection information to BaileysStartupService and Instance model

fmedeiros95:
- Updated BaileysStartupService to include disconnection handling logic.
- Enhanced Instance model to store disconnection status and details.
- Handled specific error codes to manage disconnection states:
  - 401: Triggered logout to handle unauthorized access.
  - 402: Recognized as temporary ban, reconnection attempts paused.
  - 403: Recognized as forbidden, reconnection attempts stopped.
  - 406: Recognized as permanent ban, reconnection attempts halted.

### DIFF
--- a/prisma/mysql-schema.prisma
+++ b/prisma/mysql-schema.prisma
@@ -47,34 +47,37 @@ enum TriggerOperator {
 }
 
 model Instance {
-  id               String                   @id @default(cuid())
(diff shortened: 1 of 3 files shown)


DavidsonGomes:
Ok!

PR #814, opened 2024-08-24
Receive data from Meta ADS ( facebook, instagram ), excluding Chatwoot that already haves.

diegomustafamarino:
Receive data from Meta ADS ( facebook, instagram ), excluding Chatwoot that already haves.


chac4l:
bom request

PR #854, opened 2024-09-03
Fixes media messages saving and uploading to S3 on WPP Biz API

paulillo:
Thought the last commits would resolve the problem, but the problem was even down the hole. This one fixes the entire thing


paulillo:
Desisti desse PR porque encontrei outras inconsistências no código.

PR #883, opened 2024-09-11
Debounce message break

vidz1979:
Using debounce message break as "\n" instead of white spaces allow Bot API to treat better these messages.


vidz1979:
removed docker-compose version as long it is obsolete (message when building docker image)

PR #918, opened 2024-09-25
Fix/validate message before sending chatwoot

judsonjuniorr:
Depends on https://github.com/EvolutionAPI/evolution-api/pull/917

With this before sending a new message to chatwoot it will validate if the source_id already exists on chatwoot and ignore it.

The suggestion is to prevent this kind of bug:
image


Richards0nd:
UP

dpaes:
up

PR #927, opened 2024-09-27
fix mexico number

hassan3kb:


DavidsonGomes:
The fix should be submitted for version 2.0

Issue #925, opened 2024-09-27
Improvements to integrate it over a custom WhatsappWeb / Internal Chat service


jesus-chacon:
### Welcome!

- [X] Yes, I have searched for similar requests on [GitHub](https://github.com/code-chat-br/whatsapp-api/issues) and found none.

### What type of feature?

Endpoint

## Proposal for Enhancements in Endpoints /chat/findChats/:instanceName and /chat/findMessages/:instanceName

I would like to propose some changes to the `/chat/findChats/:instanceName` and `/chat/findMessages/:instanceName` endpoints to improve their functionality.

1. **Unread Messages Count in /chat/findChats/:instanceName**:  
   It would be extremely useful to include a count of unread messages for each chat. This would allow us to easily identify which chats have pending messages.

2. **Last Message Preview in /chat/findChats/:instanceName**:  
   Adding a "lastMessage" field to display a preview of the most recent message would provide a more user-friendly experience and help to quickly identify the context of the conversation.

3. **Read Status in /chat/findMessages/:instanceName**:  
   In the `/chat/findMessages/:instanceName` endpoint, it would be beneficial to include a boolean flag for each message indicating whether it has been read. This would allow us to visually differentiate between read and unread portions of the conversation.

These enhancements would improve the usability of both endpoints and enhance the overall functionality for users handling multiple chats or messages.

## How Should the Feature Be Developed?

1. **Unread Messages Count in /chat/findChats/:instanceName**:
   - Introduce a new field in the response object that contains the number of unread messages per chat.
   - The backend can query the message table (or the relevant data source) to count messages with a `read: false` status associated with each chat.
   - This field can then be returned alongside other chat data in the `/findChats` endpoint.

2. **Last Message Preview in /chat/findChats/:instanceName**:
   - Add a `lastMessage` field in the response, which retrieves the most recent message from each chat.
   - The backend should fetch the most recent message based on the timestamp and include a preview (e.g., first 100 characters) in the response.

3. **Read Status in /chat/findMessages/:instanceName**:
   - Add a boolean field `isRead` to each message in the response of `/findMessages`.
   - The backend should check the read status of each message (e.g., a field like `read: true/false`) and return the appropriate value in the response.
   - This flag will help to distinguish between read and unread messages in the conversation history.

These changes should be backward-compatible to avoid breaking existing clients using the API.

### Additional Notes

[...]

PR #958, opened 2024-10-08
Added last message to fetch chats

jesus-chacon:
- Fix unreadMessages count.
- Added Added last message / null to fetch chats and sort by it.
- Improve message status update.
- Fix create / update multiples message.

Change "returns" in "message.upsert" and "message.update" to "continue" to allow the process of a list of messages.

Close [this issue](https://github.com/EvolutionAPI/evolution-api/issues/925)

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -983,8 +983,10 @@ export class BaileysStartupService extends ChannelStartupService {
         for (const received of messages) {
           if (received.message?.conversation || received.message?.extendedTextMessage?.text) {
             const text = received.message?.conversation || received.message?.extendedTextMessage?.text;
+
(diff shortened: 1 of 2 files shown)


DavidsonGomes:
Add verification if the DATABASE variable SAVE_DATA_NEW MESSAGE is marked as true, if it is not saving the messages I do not want it to make unnecessary calls to the database

jesus-chacon:
> Add verification if the DATABASE variable SAVE_DATA_NEW MESSAGE is marked as true, if it is not saving the messages I do not want it to make unnecessary calls to the database

@DavidsonGomes 
Right now the code check it before create the message.
Captura de pantalla 2024-10-10 a las 9 12 24

If you refer to lastMessage in fetch chats I only added values to the select block of the query the join with message was in the code that is the expensive part of the query. Please explain me a bit more where yo propuse tu implement the  SAVE_DATA_NEW_MESSAGE.

PR #983, opened 2024-10-15
refactor: Update content_attributes in ChatwootService

Richards0nd:
This commit refactors the code in ChatwootService to update the content_attributes when replyToIds are present. Instead of directly spreading the replyToIds object, the code now converts it to a JSON string and appends it to the data object.

Refactor the code to improve readability and maintainability.


lucianodltec:
This PR fix "TypeError: source.on is not a function" error when user replies a message with audio or any attachment.

PR #1605, opened 2025-06-17
Update Dockerfile

ToniShelby:
## Summary by Sourcery

Revise Dockerfile to improve build reproducibility, environment management, and image size

Enhancements:
- Install dos2unix and normalize Docker scripts line endings
- Introduce default build environment variables and dynamically generate .env for the builder
- Switch to npm ci and a dedicated build:docker script, then prune devDependencies to slim the image
- Reorder and consolidate config and source copying for better cache utilization
- Update final stage to set DOCKER_ENV, create an empty .env, and remove the prebuilt .env copy

Build:
- Copy tsup.config.ts earlier and adjust file copy patterns
- Replace npm install with npm ci --silent
- Invoke custom database generation script with defined env variables
- Remove devDependencies after build


dpaes:
your modifications its already old, have a new version of evolution api 2.3.0, and u cant put postgres by default, because have mysql and postgres option, even the credentials. And another thing, u are creating that thinking only in railway...

dpaes:
@sourcery-ai review

PR #1664, opened 2025-06-28
(mysql): remove out-of-order wavoipToken migration

pauloboc:
Delete prisma/mysql-migrations/1707735894523_add_wavoip_token_to_settings_table, which executes before the initial Setting table is created and breaks fresh MySQL installs.

The later migration 20250214181954_add_wavoip_token_column, line 145, already adds the column correctly, so keeping only that directory guarantees a clean deploy.


augustresende:
#1805

PR #1884, opened 2025-08-28
Feat/whatsapp/convert-to-jpeg

nestordavalos:
All images sent via the Baileys integration are now pre-processed and converted to JPEG format using the `sharp` library. This ensures better compatibility and prevents potential issues with unsupported formats.

- Images from URLs are now downloaded via axios before processing, which allows for the use of a proxy.
- The default filename and mimetype are updated to `image.jpg` and `image/jpeg` to reflect the conversion.

## Summary by Sourcery

Preprocess WhatsApp image attachments by fetching them via axios (respecting local proxy settings), converting buffers to JPEG using sharp, and updating file metadata to ensure compatibility.

New Features:
- Convert all WhatsApp image media to JPEG format using sharp before sending
- Download remote images via axios with optional proxy support for preprocessing

Enhancements:
- Set default image filename to image.jpg and mimetype to image/jpeg after conversion

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -2448,9 +2448,43 @@ export class BaileysStartupService extends ChannelStartupService {
     try {
       const type = mediaMessage.mediatype === 'ptv' ? 'video' : mediaMessage.mediatype;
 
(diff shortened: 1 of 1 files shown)


bergpinheiro:
Aguardando o pessoal da evo aceita o pull pois a situação tá bem no envio de imagens.

PR #1897, opened 2025-09-01
feat/validate video type before uploading to S3

nolramaf:
- Introduce SAVE_VIDEO in S3 config
- Added check for video messages before attempting S3 upload
- Prevents unnecessary uploads when SAVE_VIDEO is disabled


DavidsonGomes:
Please fix the lint with `npm run lint`

PR #1986, opened 2025-09-20
fix(evolutionbot):  Fixing the correct message sending method so that messages are split.

dersonbsb2022:
- Replace instance.textMessage() with sendMessageWhatsApp() method
- Enable message splitting by double line breaks (\n\n)
- Add proper delay and typing indicators between split messages
- Fix linkPreview parameter passing to base class methods
- Support linkPreview: false/true from webhook response
- Remove unnecessary debug logs for cleaner output

Fixes: EvolutionBot was not respecting splitMessages and linkPreview configurations

## 📋 Description

After adding linkpreview: false, messages were no longer being split correctly. Now it has been fixed, messages are being split, debouce time tested, linkpreview tested with true and false and working.

## Summary by Sourcery

Improve messaging reliability by switching to the proper sendMessageWhatsApp method, re-enabling message splitting with delays and typing indicators, and fixing linkPreview support while cleaning up debug logging

New Features:
- Support toggling link preview via webhook response

Bug Fixes:
- Restore message splitting behavior and ensure messages are split at double line breaks
- Fix linkPreview parameter propagation to messaging methods
- Use the correct sendMessageWhatsApp method instead of textMessage to respect splitMessages settings

Enhancements:
- Add typing indicators and delays between split messages
- Remove unnecessary debug logs for cleaner output


dersonbsb2022:
Desculpe @DavidsonGomes, acabei esquecendo de selecionar o branch develop :(

PR #2017, opened 2025-09-25
Fix Chatwoot DB Connection Instability and Implement Stale Conversation Cache Handling

Vitordotpy:
This pull request addresses two critical issues in the Chatwoot integration to improve its reliability and resilience. It introduces on-demand database connections to prevent intermittent connection failures and implements a retry mechanism to handle stale conversation data cached in Redis.

## Fixes:

*   **Prevent Intermittent DB Connection Errors**: Refactored the `ChatwootService` to initialize the PostgreSQL client on-demand (`getPgClient`) rather than at service instantiation. This resolves a race condition where the client could be created with an incorrect or default database connection string, causing sporadic `database does not exist` errors.
*   **Handle Stale Conversation Cache**: Implemented a `try-catch` mechanism within the `createMessage` and `sendData` methods. When a `404 Not Found` error is received from the Chatwoot API (indicating a stale conversation ID in the Redis cache), the system now automatically:
    1.  Deletes the invalid conversation entry from Redis.
    2.  Creates a new conversation in Chatwoot.
    3.  Retries the original message-sending operation with the new, valid conversation ID.

## Enhancements:

*   **Increased Resilience**: The integration is now more robust against temporary inconsistencies between the Redis cache and the Chatwoot database state, preventing message delivery failures.
*   **Improved Stability**: Lazily initializing the database connection ensures that the correct, instance-specific credentials are used for every operation, eliminating a significant source of production errors.

## Summary

This PR resolves critical stability issues in the Chatwoot integration by implementing on-demand database connections to fix intermittent connection errors and adding a retry mechanism that automatically handles stale conversation IDs from the Redis cache, ensuring more reliable message delivery.

## Summary by Sourcery

Fix intermittent database connection errors and add automatic retry handling for stale Chatwoot conversation IDs by lazily initializing the Postgres client and evicting and recreating conversations on 404 errors.

Bug Fixes:
- Prevent intermittent 'database does not exist' errors by initializing the PostgreSQL client on-demand.

Enhancements:
- Automatically detect and recover from stale Chatwoot conversation IDs by evicting the Redis cache entry, creating a new conversation, and retrying message operations.

### DIFF
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -905,33 +905,53 @@ export class ChatwootService {
     try {
       return await doCreateMessage(conversationId);
     } catch (error) {
-      const errorMessage = error.toString().toLowerCase();
-      const status = error.response?.status;
-      if (errorMessage.includes('not found') || status === 404) {
-        this.logger.warn(`Conversation ${conversationId} not found. Retrying...`);
-        const bodyForRetry = messageBodyForRetry || messageBody;
+      return this.handleStaleConversationError(
+        error,
+        instance,
+        conversationId,
+        messageBody,
+        messageBodyForRetry,
+        'createMessage',
+        (newConvId) => doCreateMessage(newConvId),
+      );
+    }
+  }
 
-        if (!bodyForRetry) {
-          this.logger.error('Cannot retry createMessage without a message body for context.');
-          return null;
-        }
+  private async handleStaleConversationError(
+    error: any,
+    instance: InstanceDto,
+    conversationId: number,
+    messageBody: any,
+    messageBodyForRetry: any,
+    functionName: string,
+    originalFunction: (newConversationId: number) => Promise<any>,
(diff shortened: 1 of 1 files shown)


andres99x:
@Vitordotpy what issues are you experiencing because of this bug? I'm having issue where certain instances won't send messages to Chatwoot, but Chatwoot can send messages to Whatsapp. Wondering if this is related to it and solves it.

Vitordotpy:
> @Vitordotpy what issues are you experiencing because of this bug? I'm having issue where certain instances won't send messages to Chatwoot, but Chatwoot can send messages to Whatsapp. Wondering if this is related to it and solves it.

@andres99x I'm currently experiencing race conditions when Evolution tries to access the Chatwoot database, as well as other issues caused by the Redis cache, which causes an error when Evolution tries to send a message to a contact whose conversations were previously deleted. Evolution tries to use cached conversations that don't exist and fails to send the messages to Chatwoot. My solution was to add a retry when sending messages to Chatwoot. If the Redis cache is out of date, the Chatwoot service will retry without cached data.

PR #2470, opened 2026-03-11
fix: mentionsEveryOne always active even when set to false

### LINKED ISSUE #2431: "sendMessage" to group with "menstionsEveryOne" always active even the value is "false"
### Welcome!

- [x] Yes, I have searched for similar issues on [GitHub](https://github.com/EvolutionAPI/evolution-api/issues) and found none.

### What did you do?

used "sendText" with "mentionsEveryOne=false" in the body to a whatsapp group.

### What did you expect?

send text to a whatsapp group with no mentions.

### What did you observe instead of what you expected?

Everyone was mentioned.

### Screenshots/Videos

Image

Image

### Which version of the API are you using?

2.3.7

### What is your environment?

Mac

### Other environment specifications

_No response_

### If applicable, paste the log output

_No response_

### Additional Notes

taking off "mentionsEveryOne" from the body stopped mentioning, either with the value "true" or "false". 

Using it on n8n 

/// Thanks for your time reading this!

enzowilliam:
## Summary
- Fixed bug where `mentionsEveryOne` was always active even when set to `false`
- Fixed validation schema property name mismatch (`everyOne` → `mentionsEveryOne`)

## Root Cause
The check `if (options?.mentionsEveryOne)` used truthy evaluation. When clients (like n8n) send `"false"` as a **string** instead of a boolean, JavaScript evaluates it as truthy because it's a non-empty string.

## Changes
1. **`whatsapp.baileys.service.ts`**: Changed truthy check to strict equality (`=== true`)
2. **`message.schema.ts`**: Renamed `everyOne` to `mentionsEveryOne` in all message schemas to match DTO

## Test plan
- [ ] Send message to group with `mentionsEveryOne: false` → should NOT mention anyone
- [ ] Send message to group with `mentionsEveryOne: true` → should mention everyone
- [ ] Send message to group without `mentionsEveryOne` → should NOT mention anyone

Fixes #2431

## Summary by Sourcery

Align WhatsApp group mention handling with the message DTO and prevent unintended @everyone mentions when the option is false or omitted.

Bug Fixes:
- Ensure `mentionsEveryOne` only triggers @everyone mentions when explicitly set to true, avoiding truthy evaluation of string values like "false".
- Correct validation schemas to use `mentionsEveryOne` instead of the mismatched `everyOne` property so options are validated and mapped consistently across message types.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -2372,7 +2372,7 @@ export class BaileysStartupService extends ChannelStartupService {
           throw new NotFoundException('Group not found');
         }
 
-        if (options?.mentionsEveryOne) {
+        if (options?.mentionsEveryOne === true) {
           mentions = group.participants.map((participant) => participant.id);
         } else if (options?.mentioned?.length) {
           mentions = options.mentioned.map((mention) => {
--- a/src/validate/message.schema.ts
+++ b/src/validate/message.schema.ts
@@ -77,7 +77,7 @@ export const textMessageSchema: JSONSchema7 = {
       description: 'Enter a value in milliseconds',
     },
     quoted: { ...quotedOptionsSchema },
-    everyOne: { type: 'boolean', enum: [true, false] },
+    mentionsEveryOne: { type: 'boolean', enum: [true, false] },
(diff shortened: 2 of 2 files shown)


DavidsonGomes (approved):
LGTM — strict === true check + schema field name correction.

PR #2515, opened 2026-04-26
fix: use correct Prisma JSON path filter in getLastMessage

### LINKED ISSUE #2495: Bug: archiveChat fails with PrismaClientValidationError on v2.3.7
## Bug Description

`POST /chat/archiveChat/{instanceName}` always returns HTTP 500 with a `PrismaClientValidationError`.

## Environment

* **Evolution API version:** 2.3.7
* **Integration:** WHATSAPP-BAILEYS
* **Database:** PostgreSQL 15

## Steps to Reproduce

1. Connect a WhatsApp instance via Baileys
2. Send a message to a contact (so a chat exists)
3. Call `POST /chat/archiveChat/{instanceName}` with:

```json
{
  "lastMessages": \[{
    "key": {
      "id": "3EB09C555B385CF03875A8",
      "fromMe": true,
      "remoteJid": "5519999999999@s.whatsapp.net"
    },
    "messageTimestamp": 1775243556
  }],
  "archive": true
}
```

4. Also tried: `PUT` method (returns 404), simplified payload `{"chat":"...","archive":true}`, and with/without `lastMessages`

## Error Response

```json
{
  "status": 500,
  "error": "Internal Server Error",
  "response": {
    "message": \[{
      "archived": false,
      "message": \[
        "An error occurred while archiving the chat. Open a calling.",
        "PrismaClientValidationError: Invalid `this.prismaRepository.contact.findMany()` invocation\\n\\nUnknown argument `remoteJid`. Available options are marked with ?"
      ]
    }]
  }
}
```

## Root Cause Analysis

The `archiveChat` method calls `G(o)` (the JID normalizer), which internally calls `whatsappNumber()`. This function executes:

```typescript
this.prismaRepository.contact.findMany({
  where: {
    instanceId: this.instanceId,
    remoteJid: { in: users.map(({ jid }) => jid) }
  }
})
```

The `Contact` Prisma model does not expose `remoteJid` as a direct filterable field in the `where` clause at line 275 of the compiled `main.js`. The error `Unknown argument "remoteJid"` confirms a schema mismatch.

## Suggested Fix

In `whatsapp.baileys.service.ts`, the `whatsappNumber` method query should match the current Prisma schema for the `Contact` model. Either:

1. Ensure `remoteJid` is a direct scalar field on the `Contact` model (not nested in JSON)
2. Or update the query to use the correct field path if the schema changed

The `archiveChat` function itself is correct — it properly calls `this.client.chatModify()`. The bug is in the JID resolution path (`G()` → `whatsappNumber()` → `contact.findMany()`).

## Impact

* `archiveChat` is completely broken on v2.3.7
* `markChatUnread` may also be affected (same code path)
* Any function routing through `whatsappNumber()` could hit this

## Workaround

Archive only at the application level. The Baileys `chatModify` call is never reached.

octo-patch:
Fixes #2495

## Problem

`POST /chat/archiveChat/{instance}` (and `markChatUnread`) always returns HTTP 500 with a `PrismaClientValidationError` when called with a `chat` string (no `lastMessage` provided).

The `getLastMessage` helper was building the Prisma `where` clause as:

```typescript
{ key: { remoteJid: number }, instanceId: this.instance.id }
```

The `key` column is a `Json`/`JsonB` field. Prisma does **not** accept nested object syntax (`{ remoteJid: value }`) to filter inside a JSON column; it requires the explicit path form:

```typescript
{ key: { path: ['remoteJid'], equals: value } }
```

Using the wrong form caused Prisma to throw `PrismaClientValidationError: Unknown argument 'remoteJid'`.

## Solution

Changed the `where` clause in `getLastMessage` to use the correct Prisma JSON path filter:

```typescript
const where: any = { key: { path: ['remoteJid'], equals: number }, instanceId: this.instanceId };
```

This matches the pattern already used elsewhere in the codebase (e.g. lines 5060-5061).

## Testing

The fix aligns with the established Prisma JSON filtering pattern used throughout the file. `archiveChat` with `{"chat": "<jid>", "archive": true}` will now correctly resolve the last message instead of throwing a validation error.

## Summary by Sourcery

Bug Fixes:
- Correct JSON path filter in getLastMessage so chat archiving and unread-marking no longer fail with PrismaClientValidationError when only a chat JID is provided.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -3689,7 +3689,7 @@ export class BaileysStartupService extends ChannelStartupService {
   }
 
   public async getLastMessage(number: string) {
-    const where: any = { key: { remoteJid: number }, instanceId: this.instance.id };
+    const where: any = { key: { path: ['remoteJid'], equals: number }, instanceId: this.instanceId };
 
     const messages = await this.prismaRepository.message.findMany({
       where,


octo-patch:
Thanks for the review!

On the `instanceId` change: `this.instanceId` is a getter that returns `this.instance.id` (see `channel.service.ts` lines 99-100). They are semantically identical, but `this.instanceId` matches the established convention used everywhere else in `whatsapp.baileys.service.ts` and the rest of the codebase, so I changed it for consistency. No behavior change is intended.

On the `Prisma.MessageWhereInput` typing: I left `where` as `any` to keep the diff minimal and focused on the bug fix. The JSON path filter (`{ key: { path: ['remoteJid'], equals: number } }`) is correct at runtime but can be awkward to express through Prisma's generated types. Happy to follow up with proper typing in a separate PR if maintainers prefer.

DavidsonGomes (approved):
LGTM — correct Prisma JSON path filter syntax.

Issue #2027, opened 2025-09-30
[BUG] Loss of synchronization and configuration when restarting the server


orlandombaa:
### 📋 Bug Description

I'm losing the instant connection when I restart the server. I'm also losing the instance configuration (like the webhook URL).

### 🔄 Steps to Reproduce

set up the service on a virtual machine in compute engine using docker compose. Here is the configuration:

services: 
n8n: 
image: n8nio/n8n:1.110.1 
container_name: n8n_service 
restart: always 
ports: 
- "5678:5678" 
environment: 
- N8N_HOST=YOUR_PUBLIC_VM_IP 
- N8N_PORT=5678 
- N8N_PROTOCOL=http 
- N8N_SECURE_COOKIE=false 
- WEBHOOK_URL=http://YOUR_PUBLIC_VM_IP:5678/ 
- DB_TYPE=postgresdb 
- DB_POSTGRESDB_HOST=YOUR_PRIVATE_DB_IP 
- DB_POSTGRESDB_PORT=5432 
- DB_POSTGRESDB_DATABASE=${DB_POSTGRESDB_DATABASE} 
- DB_POSTGRESDB_USER=${DB_POSTGRESDB_USER} 
- DB_POSTGRESDB_PASSWORD=${DB_POSTGRESDB_PASSWORD} 
- N8N_ENCRYPTION_KEY=${N8N_ENCRYPTION_KEY} 
volumes: 
- n8n_data:/home/node/.n8n 
networks: 
- evolution-net 
depends_on: 
- evolution_api 

evolution_api: 
image: atendai/evolution-api:v2.2.3 
container_name: evolution_api_service 
restart: always 
depends_on: 
- redis 
ports: 
- "8080:8080" 
environment: 
- AUTHENTICATION_API_KEY=${AUTHENTICATION_API_KEY} 
- SERVER_URL=http://YOUR_PUBLIC_VM_IP:8080 
- SERVER_PORT=8080 
- SERVER_TYPE=http 
- DATABASE_CONNECTION_URI=postgresql://your_db_user:YourEncodedPassword@YOUR_PRIVATE_DB_IP:5432/evolution-api?schema=public 
- WEBHOOK_GLOBAL_URL=http://YOUR_PUBLIC_VM_IP:8080 
# --- PERSISTENCE AND CACHE VARIABLES --- 
- DATABASE_SAVE_DATA_INSTANCE=true 
[...]

Issue #2026, opened 2025-09-30
Sync lost after reboot


orlandombaa:
Closing due to inactivity — no comments since creation. If this still reproduces on the latest release (v2.3.7+), please open a new issue with a minimal repro and current logs.

Issue #1286, opened 2025-03-05
API retornando Connection Closed na versão 2.2.3


JonathanLopes404:
### Welcome!

- [x] Yes, I have searched for similar issues on [GitHub](https://github.com/EvolutionAPI/evolution-api/issues) and found none.

### What did you do?

Estou enviando requests para a rota chat/whatsappNumbers e para a rota message/sendText

A API às vezes ta retornando sucesso, e às vezes retorna um erro 500 com esse body:
 {"status":500,"error":"Internal Server Error","response":{"message":"Connection Closed"}} 

Quando ele dá esse erro, a instância fica com status connecting por um tempo e depois volta ao normal.
Esse erro é temporário. Aparece, depois some, depois volta denovo...

E não é problema no número conectado na instância, a gente fez o teste no whatsapp, o número tava funcionando normalmente

### What did you expect?

Resposta com status 200

### What did you observe instead of what you expected?

Resposta com erro 500 e com o body:
 {"status":500,"error":"Internal Server Error","response":{"message":"Connection Closed"}} 

### Which version of the API are you using?

2.2.3

### What is your environment?

Linux

### If applicable, paste the log output

2025-03-05T19:37:05.123087074Z [1m[31m[Evolution API][0m [1m[31m  [1m[31m v2.2.3 [1m[31m 31 [0m [1m[31m - [1m[37m Wed Mar 05 2025 16:37:05   [0m [31m[41m[1m ERROR [0m [33m[1m [unhandledRejection][0m [31m[1m [object][0m [31m  [0m
2025-03-05T19:37:05.123694181Z {
2025-03-05T19:37:05.123714481Z   origin: Promise {
2025-03-05T19:37:05.123719381Z     <rejected> Error: Timed Out
2025-03-05T19:37:05.123723681Z         at /evolution/node_modules/baileys/lib/Utils/generics.js:170:32 {
2025-03-05T19:37:05.123727581Z       data: [Object],
2025-03-05T19:37:05.123730981Z       isBoom: true,
2025-03-05T19:37:05.123734381Z       isServer: false,
2025-03-05T19:37:05.123737781Z       output: [Object]
2025-03-05T19:37:05.123741181Z     }
2025-03-05T19:37:05.123744481Z   },
2025-03-05T19:37:05.123747881Z   stderr: 2
2025-03-05T19:37:05.123751281Z } 
2025-03-05T19:37:05.123754581Z 
[...]

Issue #693, opened 2024-07-13
Endpoint `instance/restart/{$instanceName}` close the session


dnl4:
### Welcome!

- [X] Yes, I have searched for similar issues on [GitHub](https://github.com/EvolutionAPI/evolution-api/issues) and found none.

### What did you do?

Called the endpoint `instance/restart/{$instanceName}`

### What did you expect?

Restart the instance

### What did you observe instead of what you expected?

The instance restarted and tried to auto connect again and sent 28 webhooks alternating the status `connecting`, `open`, `close` before stop

### Screenshots/Videos

<img width="531" alt="image" src="https://github.com/user-attachments/assets/ece89973-641a-4036-9fe8-52ec6b729b7f">

### Which version of the API are you using?

1.8.2

### What is your environment?

Mac

dpaes:
Estou fechando a issue pelo fato da versão mencionada não estar mais recebendo novas correções. Caso o problema persista em uma versão mais recente (v2), estaremos trabalhando para corrigir o quanto antes.

PR #2520, opened 2026-04-27
fix(instance.controller): emit remove.instance even when logout fails (zombie cleanup)

inovacaoecrescimento:
## Summary

When a Baileys WebSocket dies but the in-memory `waInstances[name]` entry still exists (a "zombie" instance), `deleteInstance()` in `src/api/controllers/instance.controller.ts` calls `await this.logout()` which throws `Connection Closed`. The outer `try/catch` then propagates a `BadRequestException` and **`eventEmitter.emit('remove.instance')` is never reached** — leaving the zombie in `waInstances` forever, only fixable by restarting the entire process.

This is the practical reason behind several long-standing reports about reconnect/sync issues:

- #693  — `instance/restart` closes the session
- #1286 — Connection Closed (v2.2.3)
- #2026 — Sync lost after reboot
- #2027 — Loss of synchronization on reboot

In production, operators today have **no per-instance recovery** when a Baileys socket dies in a way that prevents `logout()` from completing. Their only escape is `docker restart`, which kicks every connected user off the host and forces a full QR re-scan for everyone.

## Fix

Wrap the inner `await this.logout(...)` call in its own try/catch. Log the failure but proceed to the cleanup emit so `remove.instance` always fires. The in-memory entry **must** be purged regardless of whether logout completed cleanly — `cleaningUp()` (DB / Session / cache) and `delete this.waInstances[name]` both happen inside the `remove.instance` handler in `monitor.service.ts`.

```diff
       if (instance.state === 'connecting' || instance.state === 'open') {
-        await this.logout({ instanceName });
+        try {
+          await this.logout({ instanceName });
+        } catch (logoutError) {
+          this.logger.warn(
+            `[ZOMBIE-CLEANUP] logout failed for "${instanceName}" (likely zombie socket): ${logoutError?.toString?.() || logoutError}. Proceeding with cleanup.`,
+          );
+        }
       }
```

12 added, 1 changed line. No public API change. `DELETE /instance/:name` becomes idempotent against zombies: a caller can always recover a single instance without affecting other instances on the same host.

## Why this is safe

`remove.instance` is the canonical, internal cleanup path. The handler in `monitor.service.ts` already runs `cleaningUp()` (deletes the `Session` row, sets `connectionStatus='close'` in `Instance`, `rmSync` of the on-disk session dir, clears Redis cache, removes provider session). All of that is independent of whether `logout()` completed gracefully on the WhatsApp side. If the socket is truly dead, talking to WhatsApp wasn't going to happen anyway — what we need is the local cleanup, and this fix unblocks it.

## Tested

Patch deployed in production at Rigarr (Brazilian distributor, 14 active vendor instances, ~25k msgs/day) via Docker image overlay on v2.3.7:

- Before: a single dead Baileys socket forced `docker restart evo2_api` (~30s downtime, all 14 vendors needed to re-scan QR).
- After: `DELETE /instance/:name` returns `{"status":"SUCCESS","response":{"message":"Instance deleted"}}` even when the underlying socket is dead. Other vendors stay connected.

Smoke test before submitting: applied patch on `2.3.7` clone, ran `npm ci && npm run build`, no compile errors. Container boots cleanly. Tested DELETE on a known-zombie instance (`pedro-rca340-sp`, ~115h idle) — instance was correctly removed from memory (`Instance "X" - REMOVED` log line) and from `Instance` table.

## Related context

We're maintaining a downstream Docker overlay (`ghcr.io/inovacaoecrescimento/evolution-api-rigarr:2.3.7-zombiefix-1`) until this lands upstream. We'd love to drop the overlay and go back to using the official image — happy to address review feedback or split the change differently if useful.

Refs:
- ADR (Portuguese): https://github.com/inovacaoecrescimento/command-center/blob/main/decisions/016-zombie-recovery-evolution-patch.md
- Build & overlay repo: https://github.com/inovacaoecrescimento/evolution-api-rigarr

Signed-off-by: Bruno Cavalcante Sgarbi

## Summary by Sourcery

Bug Fixes:
- Prevent zombie instances from remaining in memory by continuing deletion and cleanup even if the logout operation throws due to a dead socket.

### DIFF
--- a/src/api/controllers/instance.controller.ts
+++ b/src/api/controllers/instance.controller.ts
@@ -456,7 +456,18 @@ export class InstanceController {
       if (this.configService.get<Chatwoot>('CHATWOOT').ENABLED) waInstances?.clearCacheChatwoot();
 
       if (instance.state === 'connecting' || instance.state === 'open') {
-        await this.logout({ instanceName });
+        try {
+          await this.logout({ instanceName });
+        } catch (logoutError) {
+          // RIGARR PATCH: zombie instance cleanup.
+          // When a Baileys socket is dead but waInstances[name] still exists,
+          // logout() throws "Connection Closed". Without this catch, the
+          // remove.instance emit below never runs, leaving the zombie in memory
+          // forever (only fixable by restarting the entire container).
+          this.logger.warn(
+            `[ZOMBIE-CLEANUP] logout failed for "${instanceName}" (likely zombie socket): ${logoutError?.toString?.() || logoutError}. Proceeding with cleanup.`,
+          );
+        }
       }
 
       try {
--- a/src/api/controllers/instance.controller.ts
+++ b/src/api/controllers/instance.controller.ts
@@ -458,15 +458,18 @@ export class InstanceController {
       if (instance.state === 'connecting' || instance.state === 'open') {
         try {
           await this.logout({ instanceName });
-        } catch (logoutError) {
-          // RIGARR PATCH: zombie instance cleanup.
-          // When a Baileys socket is dead but waInstances[name] still exists,
-          // logout() throws "Connection Closed". Without this catch, the
-          // remove.instance emit below never runs, leaving the zombie in memory
-          // forever (only fixable by restarting the entire container).
-          this.logger.warn(
-            `[ZOMBIE-CLEANUP] logout failed for "${instanceName}" (likely zombie socket): ${logoutError?.toString?.() || logoutError}. Proceeding with cleanup.`,
-          );
+        } catch (error) {
+          // logout can throw "Connection Closed" when the underlying Baileys
+          // socket is already dead but waInstances[name] still exists. We
+          // must continue to the remove.instance emit below — that is the
+          // only path that purges the in-memory entry and runs cleaningUp().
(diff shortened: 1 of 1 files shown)


inovacaoecrescimento:
Thanks for the review @sourcery-ai. Pushed a follow-up commit (`93b9081a`) addressing both points:

1. **Vendor-neutral language** — dropped the `[ZOMBIE-CLEANUP]` prefix and the `RIGARR PATCH` marker from the comment. The block now just describes the bug in upstream-friendly terms ("logout can throw 'Connection Closed' when the underlying Baileys socket is already dead but waInstances[name] still exists").

2. **Preserve diagnostic detail** — switched to structured object logging (`this.logger.warn({ message, instanceName, error })`) following the same pattern used in `monitor.service.ts` `'no.connection'` handler. The full error object reaches the logger now, so stack/cause aren't lost.

Diff: https://github.com/EvolutionAPI/evolution-api/pull/2520/commits/93b9081a

Happy to squash to one commit before merge if that's the project preference.

DavidsonGomes (approved):
LGTM — defensive try/catch ensures cleanup runs even when Baileys logout fails.

PR #2531, opened 2026-05-06
chore: bump baileys to 7.0.0-rc10 + tsx 4.21 with whatsapp-rust-bridge patch

DavidsonGomes:
## Summary
- Bumps `baileys` from `7.0.0-rc.9` to `7.0.0-rc10` and `tsx` from `^4.20.5` to `^4.21.0`.
- Adds a pnpm patch for `whatsapp-rust-bridge@0.5.3` that adds `main`, `require`, and `default` to the `exports` field so the package can be resolved from CJS projects (this repo has no `"type": "module"`).
- Migrates lockfile management to pnpm (deletes `package-lock.json`, adds `pnpm-workspace.yaml` so the patch is picked up).

## Why
Running `npm run dev:server` on `develop` was failing with:
```
ERR_PACKAGE_PATH_NOT_EXPORTED: No "exports" main defined in node_modules/whatsapp-rust-bridge/package.json
```
The bridge (a transitive dependency pulled in by baileys rc10) only declares `exports.import` + `exports.types`, with no `require`/`default`. Because evolution-api is a CJS project, both Node and tsx resolve through the CJS loader and fail.

The patch adds the missing fields without changing runtime behavior — the same `dist/index.js` is served for every condition.

## Test plan
- [x] `pnpm install` applies the patch cleanly
- [x] `pnpm run dev:server` boots the API on port 8080 and connects to WhatsApp
- [ ] CI green
- [ ] Verify other devs can install with pnpm and run dev:server

## Notes for reviewers
- Lockfile switched from npm to pnpm. If the team prefers staying on npm, this PR would need to revert the lockfile change and find another resolution mechanism (npm doesn't have a built-in patch system).
- Upstream fix should be requested at https://github.com/jlucaso1/whatsapp-rust-bridge — the `exports` field is missing `default`.

## Summary by Sourcery

Update WhatsApp integration dependencies and adopt pnpm-based patching for a transitive bridge package.

Bug Fixes:
- Ensure whatsapp-rust-bridge can be resolved correctly in CommonJS projects by applying a pnpm patch to its exports configuration.

Enhancements:
- Upgrade baileys to 7.0.0-rc10 and tsx to ^4.21.0.

Build:
- Switch lockfile management to pnpm by adding a pnpm workspace configuration and removing the npm package-lock file.


DavidsonGomes:
Reabrindo com nome de branch correto refletindo o bump do baileys.

PR #2546, opened 2026-05-17
fix: trim whitespace from instanceName on creation

### LINKED ISSUE #2543: UI fails to delete instance when `instanceName` contains trailing whitespace (404 "instance does not exist")
### Welcome

- [x] Yes, I have searched for similar issues on GitHub and found none.

### What did you do?

We have an Evolution API v2.3.7 deployment (Docker Swarm + PostgreSQL) with multiple WhatsApp instances. One instance was created with the name `"my instance "` — with a trailing space at the end (likely from a copy-paste during creation). The trailing space is invisible in the UI but persisted in the database.

From that point on, **every operation that targets the instance by name fails with 404**, including:

- `DELETE /instance/delete/{instanceName}` via API
- `DELETE /instance/logout/{instanceName}` via API
- Clicking the **Delete** button in the Evolution Manager UI

The error returned is always:

```json
{
  "status": 404,
  "error": "Not Found",
  "response": { "message": ["The \"my instance\" instance does not exist"] }
}
```

Meanwhile, `GET /instance/fetchInstances` **does** return the instance correctly, and the UI lists and displays it normally — which makes the bug very confusing to diagnose.

### Root cause (confirmed via DB inspection)

Querying the database directly:

```sql
SELECT id, name, length(name) AS len, encode(name::bytea, 'hex') AS hex
FROM "Instance"
WHERE name ILIKE '%carv%';
```

Result:

```
                  id                  |    name    | len |          hex
--------------------------------------+------------+-----+------------------------
 502f0dfa-14a7-4fdc-83f6-185e2ba90f66 | my instance |  11 | 636172762067726f7570 20
```

Decoding the hex confirms a trailing `0x20` (space) at the end of the name. The stored name is 11 chars, not 10.

### Why operations fail

- The **backend** preserves the trailing space in path parameters correctly. Calling `DELETE /instance/delete/carv%20group%20` (with the trailing `%20`) via Postman or any HTTP client **works** and returns `Instance deleted`.
- The **Evolution Manager UI**, however, appears to drop the trailing space somewhere in the frontend (likely when reading the name from a rendered DOM element, an `innerText`, or a state field that was implicitly trimmed). It ends up sending `carv%20group` (without the final `%20`), which does not match the stored name → 404.

So the backend match is strict (correct behavior), but the UI cannot delete its own instances when the stored name has trailing whitespace.

### Expected behavior

Either (or both) of:

1. **`instanceName` should be trimmed on creation.** A `.trim()` in the `createInstance` controller before persisting to the database would prevent invalid names with leading/trailing whitespace from ever being stored. This is the cleanest fix.
2. **The Manager UI should preserve the exact stored name** when issuing operations, including any trailing/leading whitespace, so its Delete/Logout/Connect buttons match what the backend expects.

Option 1 alone is enough to prevent the bug going forward. Option 2 also helps recover existing broken instances without requiring manual DB intervention.

### Steps to reproduce

1. Create an instance via API with a name that has trailing whitespace, e.g. `"test instance "` (note the trailing space):

   ```bash
   curl -X POST 'https://<server>/instance/create' \
     -H 'apikey: <GLOBAL_API_KEY>' \
     -H 'Content-Type: application/json' \
     -d '{"instanceName": "test instance ", "integration": "WHATSAPP-BAILEYS"}'
   ```

2. Confirm the instance appears in `GET /instance/fetchInstances` and in the Evolution Manager UI.
3. Try to delete it via the **Delete** button in the UI → 404 "instance does not exist".
4. Try `DELETE /instance/delete/test%20instance` → 404.
5. Try `DELETE /instance/delete/test%20instance%20` (with trailing `%20`) → **200 Instance deleted**.

### Environment

- Evolution API version: **v2.3.7**
- Client name: `evolution_api_v2_carvgroup`
- Deployment: Docker Swarm on Hetzner VPS
- Database: PostgreSQL 16.4
- Reverse proxy: Traefik

### Workaround (for anyone hitting this)

Until the fix lands, you can either:

- Call the delete endpoint manually with the exact name including trailing whitespace properly URL-encoded (`%20` at the end), or
- Run `UPDATE "Instance" SET name = TRIM(name) WHERE id = '<uuid>';` in the database, then restart the Evolution container so the in-memory monitor reloads with the corrected name.

Happy to open a PR for the `.trim()` fix if maintainers agree on the approach.

brunovigo24:
## 📋 Description

Applies `.trim()` to `instanceName` at the beginning of `createInstance` in the instance controller. This ensures that any leading or trailing whitespace is removed before the instance name is persisted to the database.

## 🔗 Related Issue

Closes #2543

## 🧪 Type of Change

## 🧪 Testing

Steps to reproduce and verify the fix:
1. Create an instance with a trailing space in the name (e.g. `"my-instance "`)
2. Confirm the instance is stored without the whitespace
3. Call `DELETE /instance/delete/my-instance` — should return 200 instead of 404

## ✅ Checklist

## 📝 Additional Notes

The root cause was that instance names with trailing spaces were stored as-is in the database, but URL path parameters used in delete/fetch operations naturally strip that whitespace — causing a name mismatch and a 404. The fix is applied at the entry point of creation so no invalid names ever reach the database.

## Summary by Sourcery

Bug Fixes:
- Trim leading and trailing whitespace from instance names before persisting them when creating instances.

### DIFF
--- a/src/api/controllers/instance.controller.ts
+++ b/src/api/controllers/instance.controller.ts
@@ -36,6 +36,8 @@ export class InstanceController {
 
   public async createInstance(instanceData: InstanceDto) {
     try {
+      instanceData.instanceName = instanceData.instanceName?.trim();
+
       const instance = channelController.init(instanceData, {
         configService: this.configService,
         eventEmitter: this.eventEmitter,


brunovigo24:
Thanks for the review! After analyzing the full request pipeline, I believe the current placement is the right approach for this specific case.

**On moving normalization earlier in the pipeline:**
The `RouterBroker.dataValidate()` is a shared base class that handles all routes across the system. Adding a trim there would affect fields in unrelated routes unnecessarily. The controller is the appropriate entry point for this business logic normalization.

**On centralizing in a shared helper:**
This would make sense if `instanceName` were persisted in multiple places, but it isn't. All other methods (`deleteInstance`, `connectToWhatsapp`, `logout`, etc.) receive the instance name via URL path parameters — not from user input. Since the fix prevents whitespace from ever being saved to the database, those methods will always receive clean names. The problem is solved at the source.

**On JSONSchema validation:**
A pattern constraint could reject names with whitespace, but it cannot transform/trim them — so it doesn't solve the root cause either.

The single-line fix in `createInstance` is intentionally minimal: it targets the only point where `instanceName` comes from user input and gets persisted, without touching shared infrastructure.

DavidsonGomes:
Atualizei a base do PR de `main` para `develop` (que é a branch ativa de desenvolvimento). Com isso o GitHub recalculou o diff e detectou conflito.

Pode rebasar em `develop` para resolver? Algo como:

```sh
git fetch upstream
git rebase upstream/develop
git push --force-with-lease
```

Depois disso, o fix de 1 linha (`instanceData.instanceName?.trim()`) está pronto para merge. Obrigado!

DavidsonGomes:
@brunovigo24 fiz um rebase na sua branch removendo commits de `main` que já estavam em `develop` (mudanças de branding `EvolutionAPI` → `evolution-foundation` em README/CONTRIBUTING/.gitmodules). Seu commit `fix: trim whitespace from instanceName on creation` foi preservado com autoria.

Force-push aplicado em `brunovigo24/evolution-api:fix/trim-instance-name-whitespace`. PR agora está MERGEABLE contra develop.

Obrigado pela contribuição!

DavidsonGomes (approved):
LGTM após rebase em develop. Fix de 1 linha (`instanceData.instanceName?.trim()`), seguro, com optional chaining defensivo.

Edge case não-bloqueante para follow-up: nome só com espaços vira string vazia e ainda passa pela validação JSONSchema. Considerar adicionar `if (!instanceData.instanceName) throw new BadRequestException('instanceName cannot be empty after trim')` em PR futuro.

PR #2552, opened 2026-05-19
fix: handle undefined ephemeralMessage for iOS PTT audio messages

### LINKED ISSUE #2550: getBase64FromMediaMessage returns ephemeralMessage error with iOS audio messages
**Version:** 2.4.0-rc2 (also tested on v2.3.7)

**Description:**
Calling POST /chat/getBase64FromMediaMessage/{instance} with an audioMessage 
received from an iOS device returns a 400 error.

**Error response:**
{"status":400,"error":"Bad Request","response":{"message":["TypeError: 
Cannot read properties of undefined (reading 'ephemeralMessage')"]}}

**Message source:** iOS (ptt: true, mimetype: audio/ogg; codecs=opus)

**Steps to reproduce:**
1. Send a voice message from an iPhone to a WhatsApp instance
2. Call POST /chat/getBase64FromMediaMessage/{instance} with the message object
3. Error is returned regardless of which fields are included in the body

**Expected behavior:** Return base64 encoded audio for transcription

**Workaround:** None found. Direct download of the URL returns encrypted .enc file.

gabrielgz0:
## Summary

Fix a crash in `POST /chat/getBase64FromMediaMessage/{instance}` when processing iOS audio messages wrapped in `ephemeralMessage` without the expected `.message` property.

## Problem

When an iPhone sends an audio message, the Baileys `messages.upsert` event delivers it wrapped in an `ephemeralMessage` container. The service iterates over known `MessageSubtype` values and unwraps them:

```typescript
for (const subtype of MessageSubtype) {
  if (msg.message[subtype]) {
    msg.message = msg.message[subtype].message; // crash
  }
}
```

iOS ephemeral messages can have `ephemeralMessage` as a truthy object **without** a standard `.message` property inside. When that happens:

1. `msg.message = undefined` (`.message` doesn't exist on the subtype)
2. Next loop iteration tries `msg.message[subtype]` -> `TypeError: Cannot read properties of undefined (reading 'ephemeralMessage')`

This also crashes when `msg.message` itself is `undefined` (e.g., passing a message object without valid content).

### Error response
```json
{"status":400,"error":"Bad Request","response":{"message":["TypeError: Cannot read properties of undefined (reading 'ephemeralMessage')"]}}
```

## Changes

**File:** `src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts`

- **Guard against falsy `msg.message`** -- returns `"Message not found"` instead of crashing
- **Safe optional chaining in subtype loop** -- `msg.message[subtype]?.message` ensures unwrapping only happens when `.message` actually exists, preventing `msg.message` from being set to `undefined`

```diff
+ if (!msg.message) {
+   throw 'Message not found';
+ }
+
  for (const subtype of MessageSubtype) {
-   if (msg.message[subtype]) {
+   if (msg.message[subtype]?.message) {
      msg.message = msg.message[subtype].message;
    }
  }
```

### Scenarios verified

| Input | Before | After |
|---|---|---|
| `ephemeralMessage` sem `.message` | TypeError crash | `"The message is not of the media type"` |
| `message: null` | TypeError crash | `"Message not found"` |
| `messageContextInfo` apenas | TypeError crash | `null` (skip) |
| Objeto vazio / string | TypeError crash | `"Message not found"` |
| `ephemeralMessage` com audio valido | works | works |
| `audioMessage` direto (DB format) | works | works |

### Checks
- [x] Lint -- `npm run lint:check` clean
- [x] TypeScript -- `tsc --noEmit` clean
- [x] Pre-commit hooks passed

Closes #2550

## Summary by Sourcery

Bug Fixes:
- Prevent crashes when processing iOS WhatsApp audio messages wrapped in ephemeral containers without a nested message object by guarding against falsy msg.message and safely unwrapping subtypes.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -4396,12 +4396,12 @@ export class BaileysStartupService extends ChannelStartupService {
 
       const msg = m?.message ? m : ((await this.getMessage(m.key, true)) as proto.IWebMessageInfo);
 
-      if (!msg) {
+      if (!msg?.message) {
         throw 'Message not found';
       }
 
       for (const subtype of MessageSubtype) {
-        if (msg.message[subtype]) {
+        if (msg.message[subtype]?.message) {
           msg.message = msg.message[subtype].message;
         }
       }


NeritonDias:
🟢 👍

Na `develop` e minimalista: optional chaining pra tratar `ephemeralMessage` undefined em áudio PTT do iOS, evitando o crash. Sem mudança estrutural. Pode seguir.

@dpaes

dpaes (approved):
Approved.

Reviewed independently against `develop`:

- `MessageSubtype` (ephemeralMessage / viewOnce / documentWithCaption) are wrappers; `TypeMediaMessage` (audioMessage etc.) are leaves — no overlap. For a non-ephemeral message, `msg.message` exists and the loop matches no subtype, so behavior is identical to before.
- For the broken iOS case (`ephemeralMessage` truthy without nested `.message`), the old code set `msg.message = undefined` and crashed on the next iteration; the `?.message` now skips it. No legitimate unwrap is lost.
- Note: the applied diff uses `!msg?.message` (more robust than the `!msg.message` described in the PR body) — it also guards a null/undefined `msg`.

Focused, additive fix (+2/-2). LGTM 👍

---

**Non-blocking suggestion for a follow-up commit** (not required to merge): the guard throws a raw string (`throw 'Message not found'`). It works because the outer catch handles it, and it matches the existing pattern in this function — but for consistency with the rest of the codebase consider throwing a proper exception class (e.g. `new BadRequestException('Message not found')`) so the error carries a status/type. Just a hygiene nit you could fold into your next commit.

PR #2561, opened 2026-05-23
fix: preserve unreadCount when inserting chats via messaging-history.set

zcag:
## Problem

When a new instance connects and Baileys fires `messaging-history.set` to bulk-import chat history, the `chatsRaw` entries are built with only `remoteJid`, `remoteLid`, `instanceId`, and `name`. The `chat.unreadCount` field — which WhatsApp's server sends in the history sync payload (`Conversation.unreadCount`, protobuf field #6) — is silently dropped.

As a result, every chat inserted through this path gets `unreadMessages = 0` (the Prisma column default), regardless of the actual unread state on the server. The `findChats` endpoint then returns `unreadCount: 0` for all historically-synced chats, even when the user has genuine unread messages.

## Root cause

The `CHATS_UPSERT` insert path mapped the field (coercing missing values to `0`):

```ts
unreadMessages: chat.unreadCount !== undefined ? chat.unreadCount : 0,
```

The `messaging-history.set` handler simply never included the equivalent mapping, so `chat.unreadCount` was dropped for every historically-synced chat.

## Fix

- Extract a shared `ChatRaw` type so the raw chat objects built for both insert paths stay in sync as fields are added.
- In the `messaging-history.set` handler, include `unreadMessages: chat.unreadCount` in the `chatsRaw.push(...)` call (previously dropped).
- Align the `CHATS_UPSERT` insert path to the same `chat.unreadCount` form: pass the value through as-is — i.e. `undefined` when the server didn't send a count — and let the Prisma column default apply, rather than coercing to `0`. This keeps both paths consistent and preserves the distinction between "server explicitly reported 0 unread" and "no unread count provided".

## Testing

1. Create a fresh instance and scan QR code
2. Let `messaging-history.set` fire during initial sync
3. Call `POST /chat/findChats/{instance}` — chats should now reflect the server-side unread counts instead of all returning `unreadCount: 0`

## Summary by Sourcery

Bug Fixes:
- Include unread message count from Conversation.unreadCount when building chats inserted during history sync, instead of defaulting all imported chats to zero unread messages.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -1116,7 +1116,7 @@ export class BaileysStartupService extends ChannelStartupService {
           contactsMapLidJid.set(contact.id, { jid });
         }
 
-        const chatsRaw: { remoteJid: string; remoteLid: string; instanceId: string; name?: string }[] = [];
+        const chatsRaw: { remoteJid: string; remoteLid: string; instanceId: string; name?: string; unreadMessages?: number }[] = [];
         const chatsRepository = new Set(
           (await this.prismaRepository.chat.findMany({ where: { instanceId: this.instanceId } })).map(
             (chat) => chat.remoteJid,
@@ -1149,7 +1149,7 @@ export class BaileysStartupService extends ChannelStartupService {
             remoteJid = chat.id;
           }
 
-          chatsRaw.push({ remoteJid, remoteLid, instanceId: this.instanceId, name: chat.name });
+          chatsRaw.push({ remoteJid, remoteLid, instanceId: this.instanceId, name: chat.name, unreadMessages: chat.unreadCount ?? 0 });
         }
 
         if (this.configService.get<Database>('DATABASE').SAVE_DATA.HISTORIC) {
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -168,6 +168,13 @@ export interface ExtendedIMessageKey extends proto.IMessageKey {
   isViewOnce?: boolean;
 }
 
+export type ChatRaw = {
+  remoteJid: string;
+  instanceId: string;
+  name?: string;
+  unreadMessages?: number;
+};
+
 const groupMetadataCache = new CacheService(new CacheEngine(configService, 'groups').getEngine());
 
 // Adicione a função getVideoDuration no início do arquivo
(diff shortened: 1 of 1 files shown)


zcag:
Pushed a commit addressing both points, thanks!

zcag:
Thanks @sourcery-ai — both points are already addressed in the current diff:

1. **Shared type:** extracted a `ChatRaw` type and applied it to both raw-chat paths — `chatsToInsert: ChatRaw[]` (the `CHATS_UPSERT` path) and `chatsRaw: (ChatRaw & { remoteLid: string })[]` (the `messaging-history.set` path) — so adding fields later stays in sync across both insert paths.

2. **DB default vs. forced `0`:** the code passes `unreadMessages: chat.unreadCount` directly rather than `?? 0`, so when the server doesn't send a count we pass `undefined` and let the Prisma column default handle it — preserving the "server explicitly said 0" vs "no value provided" distinction you flagged. I also aligned the existing `CHATS_UPSERT` path to the same form (dropping its prior `!== undefined ? … : 0`) so both paths behave identically.

The PR description has been updated to match — it previously described a `?? 0` approach that the code doesn't actually take.

NeritonDias:
🟢 👍

Na branch certa (`develop`) e focado: preserva o `unreadCount` ao inserir chats via `messaging-history.set`, com um type `ChatRaw` coerente com o que já existe. Sem reescrita. Pode seguir.

@dpaes

dpaes (approved):
Approved.

Reviewed independently against `develop`:

- **Real bug confirmed:** in `messaging-history.set` the assembled chat object never included `unreadMessages`, so the `createMany` insert always fell back to the column default `0` — every history-synced chat lost its unread count. Adding `unreadMessages: chat.unreadCount` fixes it, and the `JSON.parse(JSON.stringify(...))` round-trip drops the key when the value is `undefined`, so the column default still applies in that case (identical to before) while the real value is preserved when present.
- I also checked the change on the `chats.upsert` path (`chat.unreadCount !== undefined ? ... : 0` → `chat.unreadCount`): the `chatsToInsert` object is shared between the webhook and the DB `createMany`. When `unreadCount` is undefined, Prisma omits it and the column default `0` still applies, so the DB outcome is unchanged. Every internal consumer of `unreadMessages` (the SQL reads in `channel.service` and `updateChatUnreadMessages`) reads from the DB column, never from the webhook payload — so there is no internal dependency on the payload carrying `0`. The only observable difference is that external webhook consumers receive the key omitted instead of `0` when no count is reported, which the PR documents as intentional. No regression.
- `ChatRaw` is coherent (no name collision; compiles cleanly under `strict:false`).

LGTM 👍

---

**Non-blocking suggestion for a follow-up commit** (not required to merge): `ChatRaw` could declare the baileys-nullable fields more precisely as `name?: string | null` and `unreadMessages?: number | null` (baileys' `IConversation.name`/`unreadCount` are `string|null`/`number|null`). Harmless under the current `strict:false` config — just a typing-precision nit you could fold in next time.

PR #2608, opened 2026-06-25
fix: instance creation broken by sanitization guard, silent error swallow, and build failures

pastoriniMatheus:
## Summary

This PR fixes two cascading runtime bugs that made \`POST /instance/create\` always return 400, plus three build-time issues that broke \`npm run build\` and \`docker build\` on the v2.4.0 branch.

---

### [1] \`sanitizeUntrustedInput\` stripping \`instanceName\` from the create body

**File:** \`src/api/abstract/abstract.router.ts\`

\`PROTECTED_INSTANCE_FIELDS\` correctly blocks \`instanceName\` from being overridden via the body on routes that already receive it through the URL parameter (e.g. \`/message/sendText/:instanceName\`). However, \`POST /instance/create\` has **no URL parameter** — \`instanceName\` can only come from the body.

\`sanitizeUntrustedInput()\` was silently discarding \`instanceName\` before it reached the controller, making \`instanceData.instanceName\` undefined. Prisma 7 treats \`name: undefined\` as a missing required field and rejects \`instance.create()\` with _"Argument name is missing"_.

**Fix:** for the \`/instance/create\` route, explicitly pass \`instanceName\` through the sanitization barrier since it is a required creation input, not an untrusted override attempt.

### [2] \`saveInstance()\` swallowing the create error silently

**File:** \`src/api/services/monitor.service.ts\`

The \`try/catch\` in \`saveInstance()\` logged the error but did not rethrow it. \`createInstance()\` assumed success and continued the flow, eventually reaching \`settingsService.create()\` → \`setSettings()\` → \`setting.upsert()\`. Because the \`Instance\` row was never committed, the FK constraint on \`Setting.instanceId\` fired — surfacing as a confusing error deep inside Typebot's \`integrationSession.update()\` call in the minified bundle, pointing nowhere near the real cause.

**Fix:** rethrow after logging so the error propagates to the HTTP layer with the correct message. Additionally, \`setting.upsert()\` is now called immediately after a successful \`instance.create()\` inside \`saveInstance()\`, guaranteeing the \`Setting\` row exists before any chatbot integration can attempt to write an \`IntegrationSession\` for the new instance.

---

### [3] \`tsup.config.ts\` — \`define\` block misplaced inside \`esbuildOptions\` callback

The \`define\` object (bakes \`__LICENSE_ENDPOINT_ENCODED__\` and \`__LICENSE_ENDPOINT_XOR_KEY__\` into the bundle at compile time) was nested inside the \`esbuildOptions(options) { ... }\` callback instead of being a root-level \`defineConfig\` property. The callback was also left unclosed, producing a malformed config tsup silently ignored.

### [4] \`evohub.controlplane.router.ts\` — TypeScript type errors on \`req.params.id\`

Express types \`req.params\` properties as \`string | undefined\`. Two call sites forwarded \`req.params.id\` directly to functions expecting \`string\`, causing TypeScript strict-mode errors that aborted the build (lines 40 and 117).

### [5] \`Dockerfile\` — \`prisma.config.ts\` missing from the final stage

\`prisma.config.ts\` was only copied into the builder stage. The final image did not include it, so \`prisma migrate deploy\` at container boot failed with:

```
Error: The datasource.url property is required
```

### [6] \`.env.example\` — \`?schema=\` param causes Prisma 7 to embed wrong schema in generated client

Prisma 7 with driver adapters reads the datasource URL at \`prisma generate\` time and embeds the schema name into every generated query. With \`?schema=evolution_api\`, all SQL used \`"evolution_api"."Table"\` prefixes. The actual database schema is \`public\`, so every query failed with _table not found_. Removed \`?schema=evolution_api\` and the surrounding single-quotes (Docker Compose includes them literally in the connection string).

### [7] \`patches/\` directory — \`COPY\` directive failing on missing path

Dockerfile referenced \`COPY ./patches ./patches\` but the directory was absent from the repository, failing \`docker build\` before any application code was processed.

---

## Files changed

| File | Change |
|------|--------|
| \`src/api/abstract/abstract.router.ts\` | Pass \`instanceName\` through sanitization on \`/instance/create\` |
| \`src/api/services/monitor.service.ts\` | Rethrow errors from \`saveInstance()\`; upsert \`Setting\` immediately after \`Instance\` creation |
| \`tsup.config.ts\` | Move \`define\` block to root config; close \`esbuildOptions\` correctly |
| \`src/api/integrations/channel/evohub/evohub.controlplane.router.ts\` | Add \`as string\` cast on \`req.params.id\` (lines 40, 117) |
| \`Dockerfile\` | Copy \`prisma.config.ts\` into builder and final stages |
| \`.env.example\` | Remove \`?schema=evolution_api\` and single-quotes from \`DATABASE_CONNECTION_URI\` |
| \`patches/.gitkeep\` | Track the \`patches/\` directory required by the Dockerfile \`COPY\` directive |

## Test plan

- [ ] \`npm run build\` completes without TypeScript or tsup errors
- [ ] \`docker build\` completes without errors
- [ ] Fresh database: \`prisma migrate deploy\` runs successfully on container boot
- [ ] \`POST /instance/create\` with \`{"instanceName":"test","integration":"WHATSAPP-BAILEYS"}\` returns 201
- [ ] \`Instance\` and \`Setting\` rows exist in the database after creation
- [ ] No regression on existing instance management endpoints

## Summary by Sourcery

Fix instance creation failures caused by overzealous request sanitization and swallowed persistence errors, and restore broken build and Docker workflows.

Bug Fixes:
- Preserve instanceName in the /instance/create request payload so instances can be created successfully.
- Ensure saveInstance rethrows persistence errors and eagerly creates the related Setting record to avoid foreign key violations in downstream integrations.
- Adjust EvoHub control plane routes to treat req.params.id as a required string, resolving TypeScript strict-mode routing errors.
- Include prisma.config.ts in both builder and final Docker image stages so Prisma can read the datasource configuration at runtime.
- Update example environment configuration to avoid embedding an incorrect Prisma schema in the generated client.
- Add a tracked patches directory to satisfy the Dockerfile COPY directive during docker build.

Build:
- Move tsup define configuration to the root defineConfig options to correctly bake license constants into the bundle and fix the malformed build config.


NeritonDias:
🟢 👍

Mira a `develop` e resolve bugs reais do Prisma 7 / criação de instância. Toca arquivos de build (Dockerfile, tsup.config, .env.example), mas cada mudança é mínima e justificada pelo próprio bug (throw faltando, sanitização do `instanceName`, schema do Prisma 7). Pode seguir — só confirma se o 2º commit (auto-create database) precisa mesmo entrar junto.

@dpaes

dpaes (changes requested):
The core fix here is correct and needed — thanks for chasing it down. I'm requesting changes on **scope**, not on the fix itself.

## What's good (verified against `develop`)
- **Root cause confirmed:** the `sanitizeUntrustedInput` guard stripped `instanceName` from the body; since `POST /instance/create` has no URL param, `instanceData.instanceName` became `undefined` → Prisma 7 failed with "Argument `name` is missing", and `saveInstance` swallowed it so it later surfaced as a misleading Typebot FK error. Re-injecting **only** `instanceName` (not `instanceId`) is the right fix: it preserves the original security intent (`instanceId` stays blocked against cross-instance override) while allowing the one legitimate creation field.
- The build fixes are minimal and justified (mispositioned `tsup` `define`, `Dockerfile` referencing a non-existent `./patches`, `prisma.config.ts` needed in the final stage, `.env.example` quoting/`?schema=`). All good.

## Requested change: split out the 2nd commit
The second commit (**`auto-create database on boot`** — `create_database.js` + the `deploy_database.sh` hook) is **separable scope**: it's an additive ops feature, not required for the instance-creation fix or the build fixes. It's well-guarded (non-fatal / `exit 0`, validates `dbName` against `/^[\w-]+$/` before the DDL, idempotent, 10s timeout), so it's low-risk — but it mixes an urgent bug-fix with an ops enhancement. Please pull it into its own PR and keep this one focused on the `/instance/create` fix + build fixes. (Minor, for that separate PR: `connection.execute` runs a prepared statement for the MySQL `CREATE DATABASE` DDL — `.query` would be more idiomatic there.)

## Low nit (optional, can fold into the focused PR)
- The new `setting.upsert` in `saveInstance` is redundant with the `setSettings` upsert that runs right after (`channel.service.ts:160`) on the same row. It's idempotent and the later one overwrites with the real values, so it's harmless — just an extra write. Fine to keep if it's there to guarantee the `Setting` row before chatbot integrations.

## Note on the red CI (not caused by this PR)
For whoever looks at the failing checks: `check-lint-and-build` fails at `npx prisma generate` with `PrismaConfigEnvError: Cannot resolve environment variable DATABASE_CONNECTION_URI` — the workflow doesn't populate that env, which `prisma.config.ts` requires. **`develop` itself is already red** for the same reason, so this is pre-existing CI infra, not a defect introduced by this PR. The `BLOCKED` merge state comes from those required checks, not from a code problem here.

Once the auto-create-db commit is split out, happy to approve the focused fix. 👍

dpaes (approved):
Approving — and updating my earlier review.

I previously requested splitting the auto-create-database commit into its own PR. After a deeper pass I'm fine merging it bundled here: the ops feature is well-guarded (non-fatal `exit 0`, `dbName` validated against `/^[\w-]+$/` before the DDL, idempotent) and low-risk. Keeping the cluster together so the team is unblocked.

**Verified, good to merge:**
- **Core fix (instanceName sanitization + saveInstance rethrow):** root cause confirmed. The `throw error` has no blast radius — `saveInstance()`'s only caller is `createInstance()` (`instance.controller.ts:64`), which already wraps it in try/catch with cleanup; boot (`loadInstance`) never calls it, so startup resilience is unaffected.
- **Build fixes (tsup `define`, `as string` casts, Dockerfile `prisma.config.ts` + `patches/.gitkeep`, `.env.example`):** minimal and justified.

**Non-blocking follow-ups (mirrored on the Linear cards):**
- The `setting.upsert` in `saveInstance` is actually load-bearing, not a redundant write — `setSettings` does NOT run automatically after create (only on an explicit settings API call), so this upsert is what guarantees the default `Setting` row exists before chatbot integrations touch `IntegrationSession`. Recommend sourcing the defaults from the canonical place to avoid drift (it omits `msgCall`/`wavoipToken`, which are nullable so no break).
- `create_database.js`: `psql_bouncer` falls into the Postgres branch and the admin connection to `/postgres` is rejected by PgBouncer (session mode) — non-fatal (`exit 0`) but fails silently; worth handling that provider explicitly. Minor: `.query()` instead of `.execute()` for the MySQL DDL.

**CI note:** `check-lint-and-build` is red at `npx prisma generate` (`DATABASE_CONNECTION_URI` not populated) — pre-existing on `develop` (same failure on the 2026-06-15 push), not introduced here.

PR #2612, opened 2026-06-27
fix(baileys): preserve sessions on transient disconnects

SmartNuvem:
## Summary

This fixes a regression introduced in 2.4.0-rc2 where transient disconnects
(e.g. DisconnectReason 408) were treated as permanent session closures.

## Changes

- Restore automatic reconnect for transient disconnects (408).
- Prevent instances from being persisted in `close` state after transient failures.
- Serialize `connection.update` and `creds.update` to avoid race conditions.
- Prevent concurrent reconnect timers.
- Improve reconnect/disconnect logging.

## Validation

Tested on a production-like environment:

- Evolution API 2.4.0-rc2
- PostgreSQL
- Redis
- Cloudflare reverse proxy

Before this patch, instances disconnected within the same day or the following day and remained in `close`.

After applying this patch, the instance has remained connected beyond the period where the issue consistently occurred, without requiring QR reconnection.

No database schema, licensing, Redis, or authentication behavior was modified.


dpaes:
we do not accept anything to main, only to develop @SmartNuvem

SmartNuvem:
Thanks for the clarification! I've changed the base branch from main to
develop. Please let me know if anything else needs to be adjusted.

Em sex., 3 de jul. de 2026, 13:13, Daniel Paes ***@***.***>
escreveu:

PR #2622, opened 2026-07-02
fix: mark Chatwoot messages failed on WhatsApp error

gobeyondpty:
## Summary
- update the original Evolution message to ERROR when Baileys emits an outbound WhatsApp ERROR update
- mark the linked Chatwoot message as failed (status 3) so the agent sees the failed send instead of an apparently sent message
- store diagnostic metadata in Chatwoot additional_attributes under evolution_error, including status, WhatsApp stub code, key id, remote jid, and timestamp

## Why
Evolution already stores outbound WhatsApp ERROR updates in MessageUpdate, but the linked original Message and Chatwoot message can remain in their previous state. That makes async WhatsApp send failures invisible in Chatwoot.

We observed this with WhatsApp ERROR status 0 / code 463: the MessageUpdate row existed, but Chatwoot still showed the outbound message as sent. This patch propagates that async failure to the user-visible Chatwoot message.

## Validation
- git diff --check

I attempted npm ci --ignore-scripts to run the project checks, but dependency installation failed in this environment with npm ECONNRESET before tsc/eslint were available.

## Summary by Sourcery

Propagate outbound WhatsApp ERROR updates to the corresponding Chatwoot message so failed sends are visible to agents.

Bug Fixes:
- Update Evolution message status to ERROR when Baileys reports an outbound WhatsApp ERROR.
- Mark the associated Chatwoot message as failed when an outbound WhatsApp message errors, preventing it from appearing as successfully sent.

Enhancements:
- Store WhatsApp error diagnostics in Chatwoot message additional_attributes under evolution_error for easier troubleshooting.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -1669,8 +1669,24 @@ export class BaileysStartupService extends ChannelStartupService {
             continue;
           }
 
-          if (findMessage && update.status !== undefined && status[update.status] !== findMessage.status) {
-            if (!key.fromMe && key.remoteJid) {
+          const updateStatus = status[update.status] ?? 'SERVER_ACK';
+
+          if (findMessage && update.status !== undefined) {
+            if (key.fromMe && updateStatus === 'ERROR') {
+              await this.prismaRepository.message.update({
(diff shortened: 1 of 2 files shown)


dpaes:
we do not accept anything to main, only to develop @gobeyondpty

Issue #2618, opened 2026-06-30
WhatsApp accounts requiring Passkey can no longer connect


ArturGPJ:
### Welcome!

- [x] Yes, I have searched for similar requests on [GitHub](https://github.com/code-chat-br/whatsapp-api/issues) and found none.

### What type of feature?

Integration

### What is the motivation for the request?

One of my WhatsApp accounts has started requiring a Passkey to link new devices. There is no option to disable it.

When I try to connect through the official WhatsApp Web, after scanning the QR Code it requires creating/using a Passkey before completing the login.

<img width="1037" height="563" alt="Image" src="https://github.com/user-attachments/assets/2b64d292-b067-463c-ac0f-6ec233fdc409" />

However, with Evolution API (v2.4.0), after scanning the QR Code nothing happens. The Passkey prompt never appears, and after a while Evolution simply generates a new QR Code.

Logs:

connection: connecting
hasQr: true
qrcodeCount: 1

(scan QR Code)

...nothing happens...

hasQr: true
qrcodeCount: 2

It looks like WhatsApp is waiting for the new Passkey authentication flow, but Evolution/Baileys never reaches that step.

Environment

Evolution API: 2.4.0
WhatsApp: Latest version
Issue started only after this WhatsApp account began requiring Passkey.
Other WhatsApp accounts that do not require Passkey still connect normally.

Is this a known issue? Is there already support planned for the new Passkey authentication flow?

melissagobatti:
<img width="720" height="1280" alt="Image" src="https://github.com/user-attachments/assets/835c80fc-af07-4b27-ae24-d4ab0d1310b7" />
<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/6a0a7f03-66e9-497e-bddd-ecd39aeca3f5" />


I’m experiencing the same issue here, but on Evolution API version 2.3.7.

After scanning the QR code, nothing happens, the passkey step never appears, and after some time a new QR code is generated again.

This started only for a WhatsApp account that now requires a passkey to link new devices. Other accounts that do not require a passkey are still connecting normally.

So it does not seem to be limited to version 2.4.0. It also affects 2.3.7 in my case.

pontedigital1:
I am experiencing the same problem in version 2.3.7.

dionathanbatista:
Estou enfrentando o mesmo problema 

[...]

PR #2628, opened 2026-07-07
feat(baileys): add passkey companion-linking ceremony

mateus2001ferreira:
## Description

Adds opt-in support for WhatsApp's passkey (Shortcake/CRSC) companion-linking flow, which some numbers now require when linking a device. Implemented on top of Baileys' public API — no fork or patch of Baileys needed.

How it works:
- A browser helper extension performs the WebAuthn assertion on the web.whatsapp.com origin.
- `PasskeyCeremony` handles the server's `passkey_prologue_request` / `crsc_continuation` notifications and replies with the matching IQs (`passkey_prologue`, `companion_nonce`, `encrypted_pairing_request`).
- Public endpoints `GET/POST /passkey-ceremony/:token[/response|/confirm]` let the extension read the challenge and submit the assertion; the current stage and open URL are exposed via `connectionState`.

Gated behind `PASSKEY_CEREMONY_ENABLED=true` (default off), since a malformed IQ can get a number banned. Also requires:
- `PASSKEY_PUBLIC_URL` — API base reachable by the browser running WhatsApp Web
- `CORS_ORIGIN` allowing `*` or `https://web.whatsapp.com`

Crypto is an independent reimplementation of whatsmeow's `pair-passkey.go` (MPL-2.0); no code copied. No dependency changes (works on the current Baileys 7.0.0-rc.9).

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## Checklist
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published

## Additional Notes

The feature ships a small browser helper extension (`public/passkey-helper.zip`) that performs the WebAuthn assertion. Happy to discuss whether shipping it in-repo, as source, or as a separate distribution is preferable. Also glad to add the two env vars to `.env.example` if the approach is accepted.


j4mb0re2:
## ✅ Passkey linking working with 1Password after additional fixes

Hi!

I would like to share the results of my tests with the new Passkey Companion Linking implementation from this PR.

After some debugging and modifications, I was able to successfully connect my existing Evolution API instance to WhatsApp using a passkey stored in **1Password**.

> **Important:** I am not a programmer/developer.
>
> I use Evolution API for my own project. I used **ChatGPT** to analyze the Evolution API source code, this PR, the Evolution GO Passkey Helper extension and my server logs. ChatGPT also helped me create the modified Chrome extension and the patches described below.

---

## My original problem

My Evolution API installation was working normally before WhatsApp started requiring passkey authentication during companion-device linking.

The QR Code was generated normally.

After scanning it with WhatsApp, however, WhatsApp requested passkey/device authentication and the Evolution API instance never completed the connection.

I was originally using:

- Evolution API `v2.3.6`
- Docker / Docker Swarm
- Existing PostgreSQL database
- Existing integrations already built around Evolution API

Because my project already depends on the original Evolution API database structure and integrations, migrating everything to Evolution GO would require significant changes.

---

## Testing Evolution GO first

I installed Evolution GO to understand how the new authentication flow was supposed to work.

Evolution GO detected that WhatsApp required a passkey and generated a URL similar to:

`https://web.whatsapp.com/#wapk=...`

It also instructed me to install the official **Evolution Passkey Helper** Chrome extension.

However, I had problems completing the authentication with my setup and 1Password.

This gave us an important reference for how the passkey flow was intended to work.

---

## Creating an alternative Passkey Helper

Using the Evolution GO Passkey Helper as a reference, I asked ChatGPT to analyze how it worked and create an alternative helper specifically for my Evolution API installation.

The architecture that finally worked was:

- Chrome Manifest V3 extension
- Background Service Worker
- Content Script
- A script running in Chrome's **MAIN world** on `web.whatsapp.com`
- WebAuthn executed directly in the WhatsApp Web page context
- `navigator.credentials.get()` used from the MAIN world
- Communication between the extension and the Evolution API passkey ceremony endpoints

This was important because it allowed the WebAuthn request to be correctly detected by the **1Password Chrome extension**.

After this change, 1Password opened normally and allowed me to authenticate using my existing WhatsApp passkey.

---

## Evolution Manager does not display the Passkey URL

There is also an important UI problem.

When I perform the pairing through the Evolution Manager web interface, the QR Code appears normally.

After scanning the QR Code and WhatsApp requesting the passkey, **the Evolution Manager does not show the URL required to continue the authentication**.

Nothing appears in the web interface telling the user to open the Passkey Helper.

The URL only appeared in the Evolution API server logs:

`[Passkey] prologue_request received. Open in the browser helper: https://web.whatsapp.com/#wapk=...`

Because of this, I had to:

1. Open the Docker/server logs
2. Find the `prologue_request`
3. Copy the `https://web.whatsapp.com/#wapk=...` URL
4. Paste it manually into Chrome
5. Let the helper extension process the ceremony
6. Authenticate with 1Password

It would be very useful if **Evolution Manager automatically displayed a button/link using `passkeyOpenUrl` whenever a passkey ceremony is requested**.

For a normal user who does not have access to the server terminal, the current flow is very difficult to discover.

---

## First problem after successful 1Password authentication

After getting WebAuthn/1Password authentication working, Evolution API successfully progressed further.

The helper displayed the confirmation code.

However, when confirming it, Evolution API returned:

`ERROR [PasskeyController] confirm: ceremony has no encryption key yet`

At the same time, I noticed that Evolution API continued generating/rotating QR Codes even though the passkey ceremony was already active.

For example:

`qrcodeCount: 16`

`qrcodeCount: 17`

`qrcodeCount: 18`

`qrcodeCount: 19`

The passkey ceremony had already started, but QR processing continued.

---

## `hasActiveByInstance()` already existed

While analyzing the PR source code, we noticed that the passkey ceremony store already contains:

`passkeyCeremonyStore.hasActiveByInstance(instanceId)`

However, it did not appear to be used by the normal WhatsApp QR/reconnection handling.

Our hypothesis was that the connection/QR lifecycle could change while the passkey ceremony still depended on cryptographic state kept in memory.

That would explain why the confirmation UI existed but the controller later reported:

`ceremony has no encryption key yet`

---

## Fix 1 — Stop QR rotation while Passkey Ceremony is active

We modified `connectionUpdate()` to detect an active passkey ceremony.

Conceptually, the change was:

~~~ts
const passkeyActive =
  process.env.PASSKEY_CEREMONY_ENABLED === 'true' &&
  passkeyCeremonyStore.hasActiveByInstance(this.instanceId);

if (qr && passkeyActive) {
  this.logger.log(
    '[Passkey] ceremony active - ignoring QR rotation to preserve cryptographic state',
  );
}

if (qr && !passkeyActive) {
  // existing QR processing
}
~~~

After this modification the server correctly reported:

`[Passkey] ceremony active - ignoring QR rotation to preserve cryptographic state`

This prevented QR rotation from interfering with the active ceremony.

The previous `ceremony has no encryption key yet` behavior was no longer the main blocker.

However, pairing still did not finish.

---

## Second problem — WhatsApp/Baileys stream error 515

The next important event in the logs was:

`stream:error code="515"`

Initially, we also prevented Evolution API from reconnecting while the passkey ceremony was active.

The server reported:

`[Passkey] ceremony active - suppressing reconnect to preserve cryptographic state`

This successfully protected the ceremony state.

But there was another problem:

**the WhatsApp connection never completed.**

The browser remained processing and the device was not added to WhatsApp's linked devices.

---

## Important discovery — reconnect must depend on the Passkey stage

We then changed the logic so reconnect behavior depends on the current passkey ceremony stage.

During these stages:

- `challenge`
- `awaiting_confirmation`
- `confirmation`

the reconnect should be suppressed so that the cryptographic state is preserved.

But after:

- `confirmed`

the reconnect must be allowed.

This was the key change that finally completed the pairing.

---

## Fix 2 — Stage-aware reconnect

The working logic is currently:

~~~ts
const passkeyState =
  process.env.PASSKEY_CEREMONY_ENABLED === 'true'
    ? passkeyCeremonyStore.stateByInstance(this.instanceId)?.state.stage
    : undefined;

const passkeyInProgress =
  passkeyState === 'challenge' ||
  passkeyState === 'awaiting_confirmation' ||
  passkeyState === 'confirmation';

const shouldReconnect =
  !passkeyInProgress &&
  !codesToNotReconnect.includes(statusCode);

if (passkeyInProgress) {
  this.logger.log(
    `[Passkey] ceremony stage=${passkeyState} - suppressing reconnect to preserve cryptographic state`,
  );

  return;
}

if (passkeyState === 'confirmed') {
  this.logger.log(
    '[Passkey] ceremony confirmed - allowing reconnect to complete WhatsApp pairing',
  );
}

if (shouldReconnect) {
  await this.connectToWhatsapp(this.phoneNumber);
}
~~~

---

## 🎉 Successful result

After applying both changes, I performed a completely new pairing attempt.

The server received:

`[Passkey] prologue_request received`

I opened the generated `#wapk` URL in Chrome.

Our modified Passkey Helper triggered WebAuthn in the WhatsApp Web MAIN world.

**1Password opened correctly.**

I authenticated using my WhatsApp passkey.

The ceremony progressed.

WhatsApp/Baileys then returned:

`stream:error code="515"`

But this time Evolution API detected that the ceremony was already confirmed and logged:

`[Passkey] ceremony confirmed - allowing reconnect to complete WhatsApp pairing`

Immediately afterwards:

`CONNECTED TO WHATSAPP`

And the instance successfully connected.

I also tested the instance after pairing and **message communication worked normally**.

---

## Final working flow

The successful flow in my environment is:

~~~text
Evolution Manager generates QR
              ↓
WhatsApp scans QR
              ↓
WhatsApp requests Passkey
              ↓
passkey_prologue_request
              ↓
Evolution creates Passkey Ceremony
              ↓
#wapk URL generated
              ↓
Open URL in Chrome
              ↓
Passkey Helper executes WebAuthn
inside web.whatsapp.com MAIN world
              ↓
1Password detects the request
              ↓
Authenticate WhatsApp Passkey
              ↓
Signed WebAuthn assertion returned
to Evolution API
              ↓
Preserve Passkey cryptographic state
              ↓
Ignore QR rotation during ceremony
              ↓
Do not reconnect prematurely
              ↓
Passkey ceremony confirmed
              ↓
WhatsApp/Baileys stream 515
              ↓
Reconnect is now allowed
              ↓
CONNECTED TO WHATSAPP
              ↓
Messages working normally
~~~

---

## Suggested improvements

Based on my tests, I believe there are three areas worth reviewing.

### 1. Display `passkeyOpenUrl` in Evolution Manager

The backend already generates the URL, but I could only see it in the server logs.

When WhatsApp requests a passkey, Evolution Manager should display something like:

**"WhatsApp requires Passkey authentication — Open Passkey Helper"**

using the generated `passkeyOpenUrl`.

This would make the feature usable without requiring terminal/server access.

### 2. Protect the active Passkey Ceremony from QR rotation

When a passkey ceremony is active, new QR updates should probably not replace/interfere with the connection state required by the ceremony.

Using `passkeyCeremonyStore.hasActiveByInstance()` to detect this solved the problem in my environment.

### 3. Make reconnect Passkey-stage aware

Completely disabling reconnect while a passkey ceremony exists is not sufficient.

In my test:

`challenge / awaiting_confirmation / confirmation`

needed the cryptographic state to be preserved.

But after:

`confirmed`

Evolution needed to allow the reconnect after the WhatsApp/Baileys `515` event.

This final reconnect is what actually completed the WhatsApp connection.

---

## Test environment

- Evolution API built from PR #2628
- Evolution API reported version: `v2.3.7`
- Docker / Docker Swarm
- PostgreSQL
- Google Chrome
- 1Password browser extension
- WhatsApp account with Passkey enabled
- Custom Passkey Helper derived from studying the Evolution GO helper architecture

---

## About the custom extension

I also have the modified Chrome Passkey Helper that worked with this setup.

It was created after analyzing the Evolution GO Passkey Helper and adapting the flow so WebAuthn runs in the WhatsApp Web MAIN world and can be handled by 1Password.

I can share the extension source code if it is useful for this PR.

Before publishing it, I would prefer that someone from the project reviews the implementation, since, as mentioned above, **I am not a developer and the extension/code changes were created with ChatGPT's assistance**.

I can also provide the exact `git diff` of the Evolution API modifications that are currently working on my server.

I hope these test results help with the Passkey implementation. Thanks for working on this feature!
[evolution-go-passkey-helper-community.zip](https://github.com/user-attachments/files/31015183/evolution-go-passkey-helper-community.zip)
[evolutio-api-passkey-fix.patch](https://github.com/user-attachments/files/31015240/evolutio-api-passkey-fix.patch)
[evolution-api-passkey-helper-public.zip](https://github.com/user-attachments/files/31015320/evolution-api-passkey-helper-public.zip)

I am also attaching the generic version of the Chrome Passkey Helper used in my successful test. It contains no private server information or credentials and dynamically reads the Evolution API URL and ceremony token from the #wapk payload.
ca52f35fa29eb25dae064080885e2a9d3a1b5627727afda57599ae6c67d315cf

marcelogorutuba (changes requested):
Thanks for this, it is the only real path we have for #2618 right now. Requesting changes on two points before it can go in.

### 1. Blocking: `PasskeyRouter` is mounted inside the guarded chain with no guards

In `src/api/routes/index.router.ts` every neighbour receives `...guards`; the passkey router is the one exception, sitting in the same chain with nothing. I understand why: the helper calls these endpoints from `web.whatsapp.com`, where no `apikey` exists, and the 256-bit ceremony token is the credential. Passing `...guards` would in fact kill the feature (`instanceExistsGuard` needs `:instanceName`, `authGuard.apikey` needs a header the helper cannot have).

But "public by design" has to be explicit, not the absence of a guard on a router that lives next to guarded ones. Two concrete problems with the current shape:

- Any route added to `PasskeyRouter` later is silently public.
- The routes are mounted even when `PASSKEY_CEREMONY_ENABLED` is not `true`.

Suggested change (applied on top of your commit; `tsc --noEmit` and `eslint` pass, and the helper's `{base}/passkey-ceremony/{token}` URLs are unchanged, so the extension is unaffected):

```diff
--- a/src/api/routes/index.router.ts
+++ b/src/api/routes/index.router.ts
@@ -227,10 +227,15 @@ router
   .use('/settings', new SettingsRouter(...guards).router)
   .use('/proxy', new ProxyRouter(...guards).router)
   .use('/label', new LabelRouter(...guards).router)
-  .use('', new PasskeyRouter().router)
   .use('', new ChannelRouter(configService, ...guards).router)
   .use('', new EventRouter(configService, ...guards).router)
   .use('', new ChatbotRouter(...guards).router)
   .use('', new StorageRouter(...guards).router);
 
+// Public by design: the browser helper calls these from web.whatsapp.com, where no apikey exists.
+// The ceremony token (256-bit, 10 min TTL) is the credential. Mounted only when the feature is on.
+if (process.env.PASSKEY_CEREMONY_ENABLED === 'true') {
+  router.use('/passkey-ceremony', new PasskeyRouter().router);
+}
+
 export { HttpStatus, router };
--- a/src/api/routes/passkey.router.ts
+++ b/src/api/routes/passkey.router.ts
@@ -7,17 +7,17 @@ export class PasskeyRouter {
   public readonly router: Router = Router();
 
   constructor() {
-    this.router.get('/passkey-ceremony/:token', (req, res) => {
+    this.router.get('/:token', (req, res) => {
       const { status, body } = controller.getCeremony(req.params.token);
       res.status(status).json(body);
     });
 
-    this.router.post('/passkey-ceremony/:token/response', async (req, res) => {
+    this.router.post('/:token/response', async (req, res) => {
       const { status, body } = await controller.submitResponse(req.params.token, req.body);
       res.status(status).json(body);
     });
 
-    this.router.post('/passkey-ceremony/:token/confirm', async (req, res) => {
+    this.router.post('/:token/confirm', async (req, res) => {
       const { status, body } = await controller.confirm(req.params.token);
       res.status(status).json(body);
     });
```

### 2. Required: retarget to `develop`

The PR targets `main`, which stopped in May (Express 4, Baileys rc.9). `develop` is 143 commits ahead with Express 5 and Baileys rc13, and that is where this needs to land. I cherry-picked your commit onto `develop` to check: two trivial "both added" conflicts in `whatsapp.baileys.service.ts` (keep both sides), and one compile error after that:

```
src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts(835,35): error TS2304: Cannot find name 'session'.
```

`createClient` no longer has a local `session` on `develop`. The config still exists, so `this.configService.get<ConfigSessionPhone>('CONFIG_SESSION_PHONE').NAME` gives you the same value.

### Non-blocking note

`public/passkey-helper.zip` is a binary in the diff, so nobody can review the extension code here. Consider committing the three files (`manifest.json`, `content.js`, `README.md`) as source instead and building the zip in CI or on release.

For functional behaviour during the ceremony (QR rotation and reconnect), see the field report in the comment above from Aug 13; I have not verified those patches, but they read as plausible and worth looking at once the two points above are addressed.

mateus2001ferreira:
Thanks for the review!

**Point 1 (guard mounting):** Agreed, your suggested change is exactly right. Applied on my side: passkey routes now mount at `/passkey-ceremony` outside the guarded chain and only when `PASSKEY_CEREMONY_ENABLED === 'true'`; routes simplified to `/:token`, so the helper's `{base}/passkey-ceremony/{token}` URLs are unchanged. Patch attached.

**Point 2 (target develop + `session`):** On rc13 the `session` compile error doesn't occur, `createClient` already resolves the browser name via `this.configService.get<ConfigSessionPhone>('CONFIG_SESSION_PHONE')`, so nothing to change there.

One heads-up on mechanics: my fork's git history was rewritten at some point and no longer shares a common ancestor with evolution-foundation/evolution-api, so I can't open a clean cross-fork PR against `develop` (GitHub shows "nothing to compare"). Since you already cherry-picked onto develop, applying the attached patch on top is probably the simplest path. Happy to provide the changes in whatever format is easiest for you.
[0001-FIX-Monta-rotas-de-passkey-em-passkey-ceremony-fora-.patch](https://github.com/user-attachments/files/31752157/0001-FIX-Monta-rotas-de-passkey-em-passkey-ceremony-fora-.patch)

marcelogorutuba:
One correction to the diff in my review above. Both the PR and my suggestion mount `PasskeyRouter` after `telemetry.collectTelemetry` in `index.router.ts`. That middleware posts `req.path` to `log.evolution-api.com` (`TELEMETRY_ENABLED` defaults to `true`), and this is the only route where the credential is in the path, so every helper call ships a live ceremony token to a third party. Reproduced with `TELEMETRY_URL` pointed at a local listener: `/passkey-ceremony/<token>` arrives with the router where it is, and does not once it is mounted above the telemetry middleware.

Updated `index.router.ts` hunk (the `passkey.router.ts` hunk from the review is unchanged; `tsc --noEmit` and `eslint` still pass on `develop`):

```diff
--- a/src/api/routes/index.router.ts
+++ b/src/api/routes/index.router.ts
@@ -193,6 +193,14 @@ router.get('/assets/*splat', (req, res) => {
   }
 });
 
+// Public by design: the browser helper calls these from web.whatsapp.com, where no apikey exists.
+// The ceremony token (256-bit, 10 min TTL) is the credential and travels in the path, so this is
+// mounted BEFORE the telemetry middleware below (it reports req.path to an external endpoint).
+// Mounted only when the feature is on.
+if (process.env.PASSKEY_CEREMONY_ENABLED === 'true') {
+  router.use('/passkey-ceremony', new PasskeyRouter().router);
+}
+
 router
   .use((req, res, next) => telemetry.collectTelemetry(req, res, next))
 
@@ -227,7 +235,6 @@ router
   .use('/settings', new SettingsRouter(...guards).router)
   .use('/proxy', new ProxyRouter(...guards).router)
   .use('/label', new LabelRouter(...guards).router)
-  .use('', new PasskeyRouter().router)
   .use('', new ChannelRouter(configService, ...guards).router)
   .use('', new EventRouter(configService, ...guards).router)
   .use('', new ChatbotRouter(...guards).router)
```

mateus2001ferreira:
Good catch on the telemetry ordering, that's a real leak. Confirmed and applied your updated hunk: the passkey router is now mounted before the telemetry middleware (and still only when `PASSKEY_CEREMONY_ENABLED === 'true'`), so the ceremony token in the path never reaches `collectTelemetry`. The `passkey.router.ts` hunk is unchanged. `tsc --noEmit` passes. Updated patch attached.
[0001-FIX-Monta-rotas-de-passkey-em-passkey-ceremony-antes.patch](https://github.com/user-attachments/files/31754644/0001-FIX-Monta-rotas-de-passkey-em-passkey-ceremony-antes.patch)

Issue #2078, opened 2025-10-13
Regression: Reply/Quote breaks on v2.3.3+ and CTWA (Meta Ads) only partially works on v2.3.4. v2.3.2 is stable


gptesv:
### Welcome!

- [x] Yes, I have searched for similar issues on [GitHub](https://github.com/EvolutionAPI/evolution-api/issues) and found none.

## Summary
Since upgrading from **2.3.2**:
- **Reply / Quote** metadata stops arriving on **2.3.3** and **2.3.4** (broken).
- **CTWA (Meta Ads → WhatsApp)** first message in a **new conversation** is delivered **intermittently empty** on **2.3.2** and **2.3.4** (partial).

## Version Matrix
| EvolutionAPI | Reply / Quote | CTWA (new conversation, first message) |
| --- | --- | --- |
| **2.3.2** | ✅ Works 100% | ⚠️ Partial (sometimes empty) |
| **2.3.3** | ❌ Broken | ❓ Not validated |
| **2.3.4** | ❌ Broken | ⚠️ Partial (sometimes empty) |

### A) Reply / Quote regression
1. Run **2.3.2**, send message A, then reply to A (WhatsApp inline reply).
2. Observe webhook/payload → reply/quote metadata present (OK).
3. Upgrade to **2.3.3** (or **2.3.4**), repeat step 1.
4. **Actual**: reply/quote metadata missing or empty.  
   **Expected**: same metadata as 2.3.2.

### B) CTWA (Meta Ads → WhatsApp) intermittent empty first message
1. With **2.3.2** or **2.3.4**, trigger a new conversation from a Meta Ads **Click-to-WhatsApp** ad.
2. Check the first event delivered to our backend.
3. **Actual (intermittent)**: conversation object arrives with `messages: []` (no first message content).
4. **Expected**: first message content should be present deterministically.

### What did you expect?

Reply/Quote metadata consistently available (as in 2.3.2).

For CTWA, the first message of a new conversation is always included in the first delivery (no race/empty array).

### What did you observe instead of what you expected?

Reply/Quote missing on 2.3.3+.

CTWA first message intermittently missing (empty messages array) on 2.3.2 and 2.3.4.

### Which version of the API are you using?

Evo Api 2.3.2  - 

Because 2.3.3 and 2.3.4 doesnt work

### What is your environment?

Windows

## Environment
- EvolutionAPI: tested **2.3.2**, **2.3.3**, **2.3.4**
[...]

Issue #1872, opened 2025-08-25
[META] @lid vs @jid handling — tracking umbrella


Simplobot1:
> **Meta-issue consolidating all open reports about the `@lid` identifier replacing `@s.whatsapp.net` / JID.**
> Please do not open new `@lid`-related issues — comment here instead.

## Background

WhatsApp is rolling out `@lid` (Linked ID) identifiers for some users — it replaces the classical phone-number-based `@s.whatsapp.net` JID in certain events. Evolution API (and Baileys) currently doesn't fully translate `@lid` → phone JID in all code paths, which breaks automations, integrations and lookups.

### Webhook / event payload
- `remoteJid` arrives as `NNNNNNNNNN@lid` instead of phone-based JID, breaking downstream automations (n8n, Typebot, Chatwoot)
- `messages.upsert` / `messages.update` events started arriving with `@lid` only, no phone number
- Outgoing events carry `@lid` intermittently for the same contact

### API endpoints / database
- `GET /chat/findContacts` returns stale `@lid` identifiers after WhatsApp META update
- Replying to `@lid` JID via HTTP Request returns `400 "exists: false"`
- Error: `Unknown argument "lid". Did you mean "id"?` when sending messages
- `Chat` table stops creating new entries for `@lid` contacts (since late Sep 2025)

### Integrations
- **Typebot**: sends to `@lid` instead of JID, resulting in `BadRequestException jidOptions.exists false`
- **Chatwoot**: duplicate contacts created (one with `@lid`, one with phone JID); historical import fails for `@lid` conversations (`syncFullHistory=true` has no effect)
- **n8n / automations**: lookups based on phone number fail because webhook brings `@lid`

### Group events
- Group exit events carry `@lid` instead of JID of the leaving member

## Status

Ongoing — requires Baileys-level mapping of `@lid` → `@s.whatsapp.net`. Partial fixes have landed in various versions but the community reports symptoms persist on v2.3.7.

## For maintainers

All duplicate `@lid` issues are being closed with `state_reason: duplicate` pointing here. Use this thread as the single source of truth for LID handling across all integrations.

---

_Reorganized as a meta-issue as part of the 2026-04 issue triage. Original reporter (@Simplobot1) and all community contributors remain credited in the comment history._

albanirneves:
same here

M4rdine:
same here

NestorMiniDEV:
same here

[...]

PR #2654, opened 2026-07-23
fix(whatsapp): restore reply/quote contextInfo in prepareMessage

diegocesarmg:
## Problem

Inbound (and outbound) **text replies lose their quote/reply context**. A message that quotes another arrives at the `messages.upsert` webhook — and is persisted, and returned by `chat/findMessages` — as `messageType: "conversation"` with **no `contextInfo.quotedMessage`, no `stanzaId`, no `participant`**. There is no way for any integration (Chatwoot, n8n, custom webhooks) to link a reply to the message it quotes.

This matches the long-standing report in #2078 ("Reply/Quote breaks on 2.3.3+") and is part of the `@lid` meta #1872, but the actual cause is a small regression in `prepareMessage`.

## Root cause

In `BaileysStartupService.prepareMessage`, the message `contextInfo` is sourced from the **wrong field**:

```ts
contextInfo: this.deserializeMessageBuffers(message.message?.messageContextInfo),
```

`message.message.messageContextInfo` only carries device-list / encryption metadata (`threadId`, `deviceListMetadata`, `messageSecret`, `limitSharingV2`) — it **never** contains the quote. The reply context (`stanzaId`, `participant`, `quotedMessage`) lives on the **content node**, e.g. `message.message.extendedTextMessage.contextInfo`.

A few lines below, the `extendedTextMessage` node is flattened into `conversation` and **deleted**:

```ts
if (messageRaw.message.extendedTextMessage) {
  messageRaw.messageType = 'conversation';
  messageRaw.message.conversation = messageRaw.message.extendedTextMessage.text;
  delete messageRaw.message.extendedTextMessage;
}
```

So the quote is dropped entirely, and the `messageRaw.contextInfo.quotedMessage` handling right after it becomes **dead code** (it can never see a `quotedMessage`).

## Regression window

Every stable release sourced `contextInfo` from the content node and quotes worked:

- **2.3.2 / 2.3.3 / 2.3.7** → `contextInfo: contentMsg?.contextInfo` (with `contentMsg = message.message[getContentType(...)]`)
- **develop / homolog** → `contextInfo: message.message?.messageContextInfo` ❌

This is not a Baileys issue: 2.3.7 already runs Baileys `7.0.0-rc.9` and delivers the quote correctly. Only the source field changed.

## Fix

Restore the content-node source, with a fallback to `messageContextInfo` so nothing is lost, and reuse the already-computed content type. The `extendedTextMessage -> conversation` normalization is intentionally kept unchanged.

```ts
const contentType = getContentType(message.message);
const contentMsg = contentType ? (message.message?.[contentType] as any) : undefined;
// ...
contextInfo: this.deserializeMessageBuffers(contentMsg?.contextInfo ?? message.message?.messageContextInfo),
```

After the fix, `messages.upsert` again exposes `contextInfo.stanzaId` / `participant` / `quotedMessage` for replies, and the downstream `quotedMessage` handling in `prepareMessage` works again. `messageContextInfo` remains available under `messageRaw.message.messageContextInfo`.

Refs #2078, #1872.

## Summary by Sourcery

Bug Fixes:
- Preserve reply/quote contextInfo (stanzaId, participant, quotedMessage) for WhatsApp messages by sourcing it from the message content node instead of the top-level messageContextInfo.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -5155,6 +5155,8 @@ export class BaileysStartupService extends ChannelStartupService {
 
   private prepareMessage(message: WAMessage): Message {
     const keyAny = message.key as any;
+    const contentType = getContentType(message.message);
+    const contentMsg = contentType ? (message.message?.[contentType] as any) : undefined;
     const messageRaw: any = {
       key: {
         ...message.key,
@@ -5167,14 +5169,20 @@ export class BaileysStartupService extends ChannelStartupService {
           ? 'Você'
           : message?.participant || (message.key?.participant ? message.key.participant.split('@')[0] : null)),
       message: this.deserializeMessageBuffers({ ...message.message }),
-      messageType: getContentType(message.message),
+      messageType: contentType,
       messageTimestamp: Long.isLong(message.messageTimestamp)
         ? message.messageTimestamp.toNumber()
         : (message.messageTimestamp as number),
       source: getDevice(keyAny.id),
       instanceId: this.instanceId,
       status: status[message.status],
-      contextInfo: this.deserializeMessageBuffers(message.message?.messageContextInfo),
+      // The quote/reply context lives on the content node (e.g.
+      // extendedTextMessage.contextInfo), not on the top-level
+      // messageContextInfo (which only carries device/encryption metadata).
+      // Reading it from the content node keeps stanzaId/participant/quotedMessage
+      // available after the extendedTextMessage node is flattened below — this
+      // is how it worked through v2.3.7 and what downstream reply handling expects.
+      contextInfo: this.deserializeMessageBuffers(contentMsg?.contextInfo ?? message.message?.messageContextInfo),
     };
 
     if (!messageRaw.status && message.key.fromMe === false) {


dpaes:
@diegocesarmg have you tried in the release cantidates after 2.3.7? exists 2.4.0-rc1, 2.4.0-rc2 (both using baileys rc13)

diegocesarmg:
@dpaes thanks! Yes — I checked, and the bug is already present in **2.4.0-rc1 and 2.4.0-rc2**.

It isn't a Baileys change. `2.3.7`, `2.4.0-rc1` and `2.4.0-rc2` all pin the **same** Baileys `7.0.0-rc.9` (only `develop`/`homolog` moved to `rc13`, and it carries the same bug). So Baileys is not the variable here.

The regression is the `prepareMessage` rewrite in the `2.3.7 → 2.4.0-rc1` window, which changed where `contextInfo` is read from:

- **2.3.7** (works): `contextInfo: contentMsg?.contextInfo` — the content node, which carries the quote
- **2.4.0-rc1 / rc2 / develop** (broken): `contextInfo: message.message?.messageContextInfo` — only device/encryption metadata (`threadId`, `deviceListMetadata`, `messageSecret`), never the quote

Because `extendedTextMessage` is then flattened into `conversation` and deleted, the quote (`stanzaId` / `participant` / `quotedMessage`) is lost entirely. Same Baileys across all three — only that one line differs.

This PR just restores the content-node source (with a fallback to `messageContextInfo`), matching the 2.3.7 behavior. Happy to add a test if useful.

dpaes:
@pastoriniMatheus check this

ohmybossmobilegrooming-bit:
We hit this exact regression in production on 2.4.0-rc2 (Baileys channel) and can confirm this PR fixes it.

Since upgrading from 2.3.7 on July 12, every Click-to-WhatsApp text message lost its ad attribution: `messages.upsert` webhooks and the `Message.contextInfo` column carried only E2E metadata (`messageSecret`/`deviceListMetadata`) instead of the content node's `contextInfo`. Meta reported roughly 45 ad conversations per week for our number, and zero arrived with `externalAdReply`/`ctwaClid` for 17 days. Media messages kept their nested `contextInfo` inside the raw proto, which is how we confirmed the data was being destroyed in `prepareMessage`: the `extendedTextMessage` node is deleted right after, so for text messages the ad data is unrecoverable.

On July 29 we deployed 2.4.0-rc2 plus the same change proposed here (content-node `contextInfo` with fallback to `messageContextInfo`, wrapped in `deserializeMessageBuffers`). Within hours, both of our instances were tagging ad leads again: `externalAdReply` with `ctwaClid` and `sourceId` present in the webhook payload and in the `contextInfo` column, matching the 2.3.7 behavior.

Side note: this also revives the `messageRaw?.contextInfo?.quotedMessage` read a few lines below in `prepareMessage`, which is always undefined on rc1/rc2 since `messageContextInfo` never contains `quotedMessage`.

Would love to see this in an rc3. Happy to share more details if useful.

Issue #2500, opened 2026-04-08
[Bug]: Chatwoot integration fails to parse templateMessage - "no body message found"


FelipoArts:
Closing as duplicate of #2014 — Chatwoot integration meta-issue.

See #2014 for the full list of tracked Chatwoot issues and known workarounds.

Issue #2014, opened 2025-09-25
Bugs chatwoot (explicitando causas)


DAVIAMERICO242:
Primeiro problema:

A importação de contatos e mensagens para o chatwoot só funciona:

1 - O chatwoot for conectado antes de gerar o QRCode da instancia, e se você está conectando o QRCode pela primeira vez naquela instância.

2 - A variável de ambiente CHATWOOT_IMPORT_DATABASE_CONNECTION_URI deve usar o banco chatwoot_production não chatwoot.

Segundo problema:

Se a instância é conectada e depois desconectada e conectada novamente, então as mensagens do whatsapp não chegam no chatwoot, é preciso reiniciar o servidor da evolution para que as coisas voltem ao normal.

luizcw:
UP, isso eu já observo há várias versões da evolution, infelizmente sempre que preciso reconectar (leia-se escanear um QRCode), só funciona 1 vez após reiniciar o container da Evolution. Se desconectar por conta de algum erro, ou até mesmo manualmente, ou pra sincronizar, já era... precisa primeiro reiniciar a docker e depois ler o QRCode, ou aparece como conectado mas não aparecem as mensagens recebidas e enviadas pelo aplicativo do Whatsapp. O curioso é que continua enviando mensagem pelo chatwoot.

andres99x:
Same here! I have not been able to find the root cause yet.

vitusan:
@luizcw rapaz sorte a sua, aqui nem reiniciando nem startando do 0 com uma instalação limpa. O chatwood manda, mas ele não recebe de jeito nenhum. Novas conversas não estão sendo criadas sozinhas

DAVIAMERICO242:
[vitusan](https://github.com/vitusan) chama 85996144336 eu ajeito pra vc

HectorMouraM:
Olá pessoal, estou com um problema nessa mesma parte, mesmo fazendo o processo de conectar o chatwoot com a evo antes de ler o qrcode, aparece o erro, estou hosteando o chatwoot no portainer. Já tentei adicionar a CHATWOOT_IMPORT_DATABASE_CONNECTION_URI pelas variaveis de ambiente do portainer e nada

LouDriu:
Two related findings from debugging our own Evolution + Chatwoot (self-hosted, Baileys channel) setup, in case they help narrow down items already on this list:

**1. Message status never reaches "delivered" (only sent → read)**

[...]

PR #2666, opened 2026-07-29
fix: Chatwoot integration drops templateMessage, conversation never created

pbarbagli:
`getTypeMessage()` doesn't have a case for `templateMessage`, so when a
contact's first message arrives as a WhatsApp Business template (common
for B2B outreach sent via WhatsApp Business Platform), no content is
extracted, `eventWhatsapp` logs `"no body message found"` and returns
before ever calling `createConversation` — the conversation is silently
dropped, and only gets created later if/when a normal-type message
follows (e.g. a reply).

This adds `hydratedTemplate.hydratedContentText` extraction, mirroring
how `templateMessage` is already handled in the *other*
`getContentMessage()` fallback path a few lines below, which just isn't
used on this ingestion path.

Fixes the `templateMessage` item tracked in #2014.
Duplicate of the (closed) #2500.

Tested against a real production instance: reproduced the exact
`"no body message found"` log with a live inbound `templateMessage`,
applied this one-line fix, confirmed the conversation now gets created
correctly (as `incoming`) instead of being dropped.

## Summary by Sourcery

Bug Fixes:
- Extract content from WhatsApp templateMessage hydratedTemplate.hydratedContentText to avoid dropping conversations when the first inbound message is a template.

### DIFF
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -1858,6 +1858,7 @@ export class ChatwootService {
         msg?.message?.viewOnceMessageV2?.message?.imageMessage?.url ||
         msg?.message?.viewOnceMessageV2?.message?.videoMessage?.url ||
         msg?.message?.viewOnceMessageV2?.message?.audioMessage?.url,
+      templateMessage: msg.templateMessage?.hydratedTemplate?.hydratedContentText,
     };
 
     return types;
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -1858,7 +1858,7 @@ export class ChatwootService {
         msg?.message?.viewOnceMessageV2?.message?.imageMessage?.url ||
         msg?.message?.viewOnceMessageV2?.message?.videoMessage?.url ||
         msg?.message?.viewOnceMessageV2?.message?.audioMessage?.url,
-      templateMessage: msg.templateMessage?.hydratedTemplate?.hydratedContentText,
+      templateMessage: msg?.templateMessage?.hydratedTemplate?.hydratedContentText,
     };
 
     return types;
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -1866,6 +1866,8 @@ export class ChatwootService {
       albumMessage: msg?.albumMessage
         ? `[álbum: ${msg.albumMessage.expectedImageCount ?? 0} foto(s), ${msg.albumMessage.expectedVideoCount ?? 0} vídeo(s)]`
         : undefined,
+      pollCreationMessage: msg?.pollCreationMessage?.name,
+      groupInviteMessage: msg?.groupInviteMessage?.caption || msg?.groupInviteMessage?.groupName,
     };
 
     return types;
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -1859,6 +1859,13 @@ export class ChatwootService {
         msg?.message?.viewOnceMessageV2?.message?.videoMessage?.url ||
         msg?.message?.viewOnceMessageV2?.message?.audioMessage?.url,
       templateMessage: msg?.templateMessage?.hydratedTemplate?.hydratedContentText,
+      buttonsMessage: msg?.buttonsMessage?.contentText,
+      interactiveMessage: msg?.interactiveMessage?.body?.text,
+      // AlbumMessage carries no text (WAProto: only expectedImageCount/expectedVideoCount) —
+      // the photos/videos arrive as separate associatedChildMessage-wrapped messages.
+      albumMessage: msg?.albumMessage
+        ? `[álbum: ${msg.albumMessage.expectedImageCount ?? 0} foto(s), ${msg.albumMessage.expectedVideoCount ?? 0} vídeo(s)]`
+        : undefined,
     };
 
     return types;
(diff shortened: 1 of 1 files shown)


abckx-tech:
### Production data confirming this — and evidence the same root cause affects more types than `templateMessage`

We hit this in production and only found it by reconciling the Evolution DB against the Chatwoot DB directly. Sharing the numbers in case they help prioritise.

**Setup:** `evoapicloud/evolution-api:v2.3.7`, Chatwoot 4.13.0 (self-built image; no changes to the API-channel ingestion path), Chatwoot integration enabled, 4 instances.

**Method:** join `Message.key.id` (Evolution) against `messages.source_id` (Chatwoot, stored as `WAID:<id>`). Validated in both directions before trusting it — picked a message the join says arrived and confirmed it exists, and picked several it says are missing and searched Chatwoot by full text content. 1 false positive in 6 sampled, so treat these as upper bounds.

**Window:** 60 days, inbound only, 1:1 conversations (groups excluded).

| message type | inbound in Evolution | absent from Chatwoot | drop rate |
|---|---|---|---|
| `templateMessage` | 37 | 37 | **100%** |
| `secretEncryptedMessage` | 24 | 24 | **100%** |
| `associatedChildMessage` | 13 | 13 | **100%** |
| `buttonsMessage` | 12 | 12 | **100%** |
| `albumMessage` | 11 | 11 | **100%** |
| `interactiveMessage` | 7 | 7 | **100%** |
| `placeholderMessage` | 3 | 3 | **100%** |

`templateMessage` is **37 for 37 — not a single one produced a conversation**, which matches the `"no body message found"` early-return described here exactly.

The rest of the table is what I think is worth adding to the discussion: every type without a case in `getTypeMessage()` drops at 100%. So the one-line fix here is correct but treats one symptom of a shared shape — a type the extractor doesn't know silently ends the ingestion path instead of falling through.

Two suggestions, offered as a user rather than a maintainer:

1. Consider falling back to `getContentMessage()` (or a generic placeholder) for any unrecognised type, so an unknown type degrades to a conversation with poor content rather than no conversation at all.
2. Consider logging `"no body message found"` at `warn`/`error` with the message type included. Right now this failure is completely silent from the operator's side: no error surfaces, no queue grows, nothing retries. We ran for weeks without knowing, and the only way we detected it was cross-checking two databases by hand.

Happy to provide more detail from our dataset if useful.

pbarbagli:
> ### Production data confirming this — and evidence the same root cause affects more types than `templateMessage`
> We hit this in production and only found it by reconciling the Evolution DB against the Chatwoot DB directly. Sharing the numbers in case they help prioritise.
> 
> **Setup:** `evoapicloud/evolution-api:v2.3.7`, Chatwoot 4.13.0 (self-built image; no changes to the API-channel ingestion path), Chatwoot integration enabled, 4 instances.
> 
> **Method:** join `Message.key.id` (Evolution) against `messages.source_id` (Chatwoot, stored as `WAID:<id>`). Validated in both directions before trusting it — picked a message the join says arrived and confirmed it exists, and picked several it says are missing and searched Chatwoot by full text content. 1 false positive in 6 sampled, so treat these as upper bounds.
> 
> **Window:** 60 days, inbound only, 1:1 conversations (groups excluded).
> 
> message type	inbound in Evolution	absent from Chatwoot	drop rate
> `templateMessage`	37	37	**100%**
> `secretEncryptedMessage`	24	24	**100%**
> `associatedChildMessage`	13	13	**100%**
> `buttonsMessage`	12	12	**100%**
> `albumMessage`	11	11	**100%**
> `interactiveMessage`	7	7	**100%**
> `placeholderMessage`	3	3	**100%**
> `templateMessage` is **37 for 37 — not a single one produced a conversation**, which matches the `"no body message found"` early-return described here exactly.
> 
> The rest of the table is what I think is worth adding to the discussion: every type without a case in `getTypeMessage()` drops at 100%. So the one-line fix here is correct but treats one symptom of a shared shape — a type the extractor doesn't know silently ends the ingestion path instead of falling through.
> 
> Two suggestions, offered as a user rather than a maintainer:
> 
> 1. Consider falling back to `getContentMessage()` (or a generic placeholder) for any unrecognised type, so an unknown type degrades to a conversation with poor content rather than no conversation at all.
> 2. Consider logging `"no body message found"` at `warn`/`error` with the message type included. Right now this failure is completely silent from the operator's side: no error surfaces, no queue grows, nothing retries. We ran for weeks without knowing, and the only way we detected it was cross-checking two databases by hand.
> 
> Happy to provide more detail from our dataset if useful.

Thanks @abckx-tech for the production data — really useful. We hit the same thing on our own instance (a secretEncryptedMessage got silently dropped a few hours after we noticed this thread), so I went ahead and broadened this PR to cover the full pattern instead of just templateMessage.
Looked up each type's actual shape in Baileys' WAProto.proto before deciding how to handle it, since a blanket fallback would've been wrong for a couple of these:
buttonsMessage / interactiveMessage: these actually carry real text (contentText / body.text) that just wasn't being extracted — added real extraction, not a placeholder.
albumMessage: carries no text by design (WAProto only has expectedImageCount/expectedVideoCount — the photos arrive as separate messages). Synthesizes an informative label instead.
associatedChildMessage: it's just a FutureProofMessage envelope wrapping a real embedded message (e.g. an album's individual photo) — now unwrapped recursively instead of treated as its own unknown type.
placeholderMessage: this one is WhatsApp-internal only (MASK_LINKED_DEVICES), never real user content — explicitly skipped rather than warned/dropped, so it stops padding the "silent drop" stats.
secretEncryptedMessage: it's an encrypted edit/event payload targeting an existing message (SecretEncType.EVENT_EDIT/MESSAGE_EDIT), not new content — decrypting and applying it as an edit is a separate, bigger piece of work I didn't attempt here, so it's explicitly skipped with a clear reason instead of logging a misleading "no body" warning.
Also added pollCreationMessage and groupInviteMessage — two more common person-to-person types with no extractor.
For anything else not covered above (current or future WhatsApp message types), instead of silently dropping we now create the conversation with a [mensagem não suportada: <tipo>] placeholder and log the actual type name — so a gap like this one is visible from the logs instead of only discoverable by reconciling two databases by hand.
Tested against our own instance (tsc/eslint clean, deployed and monitoring). Happy to adjust based on maintainer feedback.

dpaes:
@pbarbagli please send to develop (we dont accept anything to main), so we can review that

pbarbagli:
@dpaes done.

PR #2671, opened 2026-08-03
feat: implement WAMonitoringService for instance lifecycle management and add WhatsApp Baileys integration support

codedalex:
## Summary by Sourcery

Improve WhatsApp Baileys instance lifecycle handling and message/contact persistence, and adjust local docker networking defaults.

Bug Fixes:
- Add a fallback mechanism to populate pushName for incoming WhatsApp messages using existing contact records when it is missing from the payload.
- Prevent duplicate WhatsApp messages from being created by checking for an existing message record before inserting a new one.
- Ensure WhatsApp instances are not left in a zombie state by forcefully closing and cleaning up any existing instance with the same name before initializing a new one.

Enhancements:
- Refine WhatsApp contact persistence so pushName is stored only for non-self contacts when available.

Deployment:
- Update docker-compose to expose the API on a different local port and simplify networking by removing the external dokploy network.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -9,12 +9,11 @@ services:
       - redis
       - evolution-postgres
     ports:
-      - "127.0.0.1:8080:8080"
+      - "127.0.0.1:8081:8080"
     volumes:
       - evolution_instances:/evolution/instances
     networks:
       - evolution-net
-      - dokploy-network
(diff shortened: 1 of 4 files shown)


dpaes:
@codedalex send to develop, so we can review that PR (we dont accept anything to main)

PR #2674, opened 2026-08-05
fix(baileys): expose delivery receipts for group messages

gersonkm:
## Problem

Delivery confirmation for **group messages** never reaches the webhook.

Individual chats confirm normally (`SERVER_ACK` → `DELIVERY_ACK` → `READ`), but group messages produce **no event at all** — even though WhatsApp does send the receipt and Baileys does receive it, one per participant.

## Root cause

Baileys routes receipts for group messages to `message-receipt.update` instead of `messages.update` (`handleReceipt` in `messages-recv.ts`):

```js
if (isJidGroup(remoteJid) || isJidStatusBroadcast(remoteJid)) {
    const updateKey = status === DELIVERY_ACK ? 'receiptTimestamp' : 'readTimestamp';
    ev.emit('message-receipt.update', ...)
}
```

Note that **delivery arrives as `receiptTimestamp`**, while read arrives as `readTimestamp`. The handler for that event consumes only `readTimestamp` and never calls `sendDataWebhook`, so the delivery signal is silently dropped.

## Evidence

Measured on 2.3.7 with `LOG_BAILEYS=debug`. A single message sent to a 3-participant group produced four receipt stanzas (identifiers redacted):

```json
{"recv":{"tag":"receipt","attrs":{"from":"<group>@g.us","id":"<msg-id>","participant":"<user-a>:11@lid"}}}
{"recv":{"tag":"receipt","attrs":{"from":"<group>@g.us","id":"<msg-id>","participant":"<user-b>:6@lid"}}}
{"recv":{"tag":"receipt","attrs":{"from":"<group>@g.us","id":"<msg-id>","participant":"<user-b>@lid"}}}
{"recv":{"tag":"receipt","attrs":{"from":"<group>@g.us","type":"read","id":"<msg-id>","participant":"<user-a>:11@lid"}}}
```

Webhook calls in the same window: **zero**.

With this patch applied, the same test produces four `MESSAGES_UPDATE` events with `participant` populated and status `DELIVERY_ACK` / `READ`.

## Approach

Emits `MESSAGES_UPDATE` using the **same flat payload** as the individual-chat branch, so that:

- no new event type is needed in the `Events` enum or in the per-integration schemas (webhook, RabbitMQ, NATS, SQS, Kafka, Pusher);
- existing consumers keep working — `participant` is already part of the status schema;
- the existing `readTimestamp` handling is left untouched.

## Notes

- Running in production since 2026-08-05, on top of 2.3.7.
- `npm run build` and `npm run lint:check` pass.
- No tests added: the repository has no coverage around this event path, and reproducing it requires a live WhatsApp session with a group.

### DIFF
--- a/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
+++ b/src/api/integrations/channel/whatsapp/whatsapp.baileys.service.ts
@@ -2163,6 +2163,19 @@ export class BaileysStartupService extends ChannelStartupService {
               const remotesJidMap: Record<string, number> = {};
 
               for (const event of payload) {
+                // Group receipts arrive here instead of `messages.update`, with
+                // delivery carried by `receiptTimestamp` and read by `readTimestamp`
+                // (Baileys sets exactly one of them per event). Only the latter was
+                // consumed, so group delivery never reached the webhook.
+                this.sendDataWebhook(Events.MESSAGES_UPDATE, {
+                  keyId: event.key.id,
+                  remoteJid: event.key.remoteJid,
+                  fromMe: event.key.fromMe,
+                  participant: event.receipt.userJid,
+                  status: typeof event.receipt.receiptTimestamp === 'number' ? 'DELIVERY_ACK' : 'READ',
+                  instanceId: this.instanceId,
+                });
+
                 if (typeof event.key.remoteJid === 'string' && typeof event.receipt.readTimestamp === 'number') {
                   remotesJidMap[event.key.remoteJid] = event.receipt.readTimestamp;
                 }


gersonkm:
Thanks for the review. I looked into all three points — one is addressed, two don't apply, and I think it's worth showing why.

### 1. `receiptTimestamp` vs `readTimestamp` — they can't coexist

Baileys builds the receipt object with a **computed key**, so exactly one of the two fields exists per event (`Socket/messages-recv.ts`):

```js
const updateKey = status === proto.WebMessageInfo.Status.DELIVERY_ACK ? 'receiptTimestamp' : 'readTimestamp';
ev.emit('message-receipt.update', ids.map(id => ({
    key: { ...key, id },
    receipt: {
        userJid: jidNormalizedUser(attrs.participant),
        [updateKey]: +attrs.t
    }
})));
```

The check in this PR is the mirror image of how Baileys populates the object, so a read receipt can't be misclassified as delivery — `receiptTimestamp` is simply absent on it.

### 2. Duplicate updates for reads

`updateMessagesReadedByTimestamp` doesn't emit anything — it runs a `$executeRaw` UPDATE and returns the affected row count. There is no `sendDataWebhook` call in that path, so a read produces exactly one webhook event (the one added here) plus the same database update as before.

### 3. Comment length — fixed

Fair point, and applied: trimmed from 9 lines to 4, keeping only what isn't obvious from the code (that group receipts land on a different event, and which field carries delivery). The longer narrative lives in the PR description, which is where future readers will look for it.

Let me know if you'd prefer the status derivation to read Baileys' `status` field explicitly instead — I kept it timestamp-based to match how the surrounding code already works, but I'm happy to change it.

royvinz:
Independent confirmation from another deployment, in case it helps move this along.

Same finding, reached from the opposite direction — we were building delivery
tracking on top of the `messages.update` webhook and could not understand why
group sends never produced anything.

**Measured on 2.3.7**, over the full `MessageUpdate` history of a live instance
(1 493 rows):

| target | SERVER_ACK | DELIVERY_ACK | READ | ERROR |
|---|---|---|---|---|
| direct  | 364 | 676 | 453 | 0 |
| group   | 0 | 0 | 0 | 12 |

Direct chats follow the full lifecycle. Groups produce **nothing** — not even
`SERVER_ACK` — across every group on the instance, over months.

The only status groups ever receive is `ERROR`, and it is not a delivery
failure: it comes from the separate `attrs.error` branch a few lines below,
whose own comment reads *"device could not display the message"*. We observed it
on messages that were demonstrably received, sent by a recipient device that had
failed to decrypt. It does not distinguish group from direct, which is why it is
the one status that reaches groups.

We also checked whether the information could be reached by configuration
instead: none of the 30 `WEBHOOK_EVENTS_*` carries `message-receipt.update`, so
there is no way to consume these receipts without this change.

One point in favour of the payload shape chosen here: `keyId` flat alongside
`status` is already what a consumer written against `messages.update` will
accept, since Evolution emits that shape elsewhere. Our own webhook handler
parses this PR's payload unchanged — no consumer-side change needed.

PR #2685, opened 2026-08-12
fix(auth-state): restore AppStateSyncKeyData.fromObject for app-state sync keys

tgiorgio:
## Problem

The three auth-state providers read the persisted app-state sync key with `proto.Message.AppStateSyncKeyData.create(value)`:

- `src/utils/use-multi-file-auth-state-prisma.ts:185`
- `src/utils/use-multi-file-auth-state-redis-db.ts:64`
- `src/utils/use-multi-file-auth-state-provider-files.ts:119`

`create()` is a bare constructor: it copies own properties and performs **no type coercion**. `fromObject()` is the one that converts, and for `bytes` fields it base64-decodes a string into a `Uint8Array`. Baileys' own reference implementation uses `fromObject()` for exactly this reason (`Utils/use-multi-file-auth-state.ts`).

The value is stored as base64 text because protobufjs `Message.toJSON()` uses `util.toJSONOptions` with `bytes: String`, so the `Uint8Array` is already a string by the time `BufferJSON.replacer` runs. Reading it back with `create()` leaves it a string, and the round trip is asymmetric.

## Consequence

`hkdf()` receives a string and evaluates `new Uint8Array("7ae+...")` → `ToIndex(NaN)` → **length 0**. Every mutation key is derived from empty key material:

```
Invalid patch mac
failed to sync state from version, removing and trying from scratch
error:1C800064:Provider routines::bad decrypt
resyncing regular from v0        <- loops forever
```

It looks fine for the first few minutes after pairing because `makeCacheableSignalKeyStore` serves the original Buffer-bearing object from its NodeCache (`SIGNAL_STORE`, 5 min TTL). After expiry the cold read returns the corrupt string — and writes it back into the cache, so it never recovers on its own.

The whole `regular` collection is affected. The way it usually surfaces: **chat labels applied on the phone stop reaching the webhook**, `labels.association` fires during pairing and never again.

A useful diagnostic: a healthy instance persists `app-state-sync-version-*` for all five collections; a broken one only ever has `critical_block`.

## Fix

`create(` → `fromObject(` at the three call sites. `grep -rn AppStateSyncKeyData src/` returns exactly those three hits.

## Regression origin

Introduced in 8830f476e898fc3d86a6678d44369a1b61a1f244 (bump to v2.3.3). v2.3.2 is clean; v2.3.3 through v2.3.7, `main` and `develop` all carry it. The three files are byte-identical across those refs, so this patch applies unchanged to all of them.

## Verification

Reproduced and fixed on a real 2.3.7 instance (Baileys 7.0.0-rc.9), with both the Redis and the file storage backends — they fail identically, which is what pointed above the storage layer in the first place.

Empirically, on the same instance:

```
create()     -> keyData is a 44-char String
fromObject() -> keyData is the correct 32-byte Buffer
```

After the change, the instance resynced `regular` from v0 to v19 with zero decrypt errors and label events resumed immediately. **Keys already persisted in the broken format are recovered as-is — no re-pairing is required**, which matters for anyone who has been running a broken instance for a while.

## Note

This is the same fix as #2593, which was closed for targeting `main`. This one targets `develop` per CONTRIBUTING.

## Summary by Sourcery

Bug Fixes:
- Fix app-state sync key deserialization by using the protobuf fromObject API so persisted keys are loaded as binary data rather than base64 strings, preventing broken HKDF derivation and state sync failures across Prisma, file, and Redis auth-state backends.

### DIFF
--- a/src/utils/use-multi-file-auth-state-prisma.ts
+++ b/src/utils/use-multi-file-auth-state-prisma.ts
@@ -182,7 +182,7 @@ export default async function useMultiFileAuthStatePrisma(
             ids.map(async (id) => {
               let value = await readData(`${type}-${id}`);
               if (type === 'app-state-sync-key' && value) {
-                value = proto.Message.AppStateSyncKeyData.create(value);
+                value = proto.Message.AppStateSyncKeyData.fromObject(value);
               }
 
               data[id] = value;
--- a/src/utils/use-multi-file-auth-state-provider-files.ts
+++ b/src/utils/use-multi-file-auth-state-provider-files.ts
@@ -116,7 +116,7 @@ export class AuthStateProvider {
               ids.map(async (id) => {
                 let value = await readData(`${type}-${id}`);
                 if (type === 'app-state-sync-key' && value) {
-                  value = proto.Message.AppStateSyncKeyData.create(value);
+                  value = proto.Message.AppStateSyncKeyData.fromObject(value);
                 }
 
                 data[id] = value;
--- a/src/utils/use-multi-file-auth-state-redis-db.ts
+++ b/src/utils/use-multi-file-auth-state-redis-db.ts
@@ -61,7 +61,7 @@ export async function useMultiFileAuthStateRedisDb(
             ids.map(async (id) => {
               let value = await readData(`${type}-${id}`);
               if (type === 'app-state-sync-key' && value) {
-                value = proto.Message.AppStateSyncKeyData.create(value);
+                value = proto.Message.AppStateSyncKeyData.fromObject(value);
               }
 
               data[id] = value;


XSERGIODM:
Independent confirmation on a production 2.3.7 instance (Baileys 7.0.0-rc.9, Redis auth
state), plus one extra symptom that may be worth adding to the issue description.

**Reproduced the asymmetry in isolation**, running the real `baileys@7.0.0-rc.9` package
against a fabricated key:

```
create({keyData:"AAAA"}).keyData      -> string "AAAA"   (44-char string for a real 32-byte key)
fromObject({keyData:"AAAA"}).keyData  -> Buffer len 3    (32-byte Buffer for a real key)
```

Full encrypt/decrypt cycle with the same fabricated key: `create()` fails with
`error:1C800064:Provider routines::bad decrypt` — the exact error string we see in production —
while `fromObject()` decrypts fine.

**The `dist` of the running container has it too**, not just the tagged source. Our image
is `:latest`, so we read `/evolution/dist/main.js` directly (485,067 chars, searched by
character offset because tsup emits multi-hundred-KB lines): all three auth-state branches
carry `AppStateSyncKeyData.create(...)` — Redis, Prisma and provider-files.

**Why a healthy-looking session hides it.** `creds` survives the Redis round trip intact
because `Buffer.prototype.toJSON()` emits `{"type":"Buffer","data":[…]}`, which
`BufferJSON.replacer` preserves. A protobuf message's `toJSON()` emits bare base64 and
destroys that marker. So the session reconnects across restarts forever while the
app-state key has been wrong **since its very first write** — which is why "it worked
yesterday" and re-pairing are both red herrings.

**An extra symptom, beyond labels: the call log.** We were chasing a different bug —
outbound calls never reaching the webhook — and it led here. With `LOG_BAILEYS=debug`, a
real outbound call produced two `server_sync` notifications for the `regular` collection,
at its start and at its end, and both died in `decodeSyncdMutations` with the same
`bad decrypt`, including the retry from v0. `SyncActionValue.callLogAction` is declared in
the same Baileys version, so the mutation is very likely the call-log record — but with
the key broken it cannot be decoded, so that stays a hypothesis on our side. It does show
the blast radius is wider than labels.

**Worth stressing for anyone triaging this: the failure is invisible at the default log
level.** `chats.js` logs `resyncing <collection> from v<n>` and `failed to sync state from
version` through `logger.info` (level 30), and `LOG_BAILEYS` defaults to `error` (50).
Grepping a normal day's logs for `bad decrypt` returns zero **whether or not the instance
is broken**. The version-key diagnostic in the PR description — a healthy instance keeps
`app-state-sync-version-*` for all five collections, a broken one only ever has
`critical_block` — is far more reliable, and it is the one we would suggest putting in the
first paragraph.

Nothing here contradicts the patch; it is the same three lines. Just more evidence that it
is the right three.

PR #2691, opened 2026-08-13
fix(meta): guard non-string mimetype before image check (fixes #2678)

### LINKED ISSUE #2678: Bug: Chatwoot → Cloud API audio fails with `TypeError: t.mimetype?.startsWith is not a function
## Environment

- Evolution API: v2.3.7
- Chatwoot: latest
- Integration: WhatsApp Business Cloud API
- Storage: IDrive E2 (S3 compatible)
- Docker deployment

## Problem

Sending an audio attachment from Chatwoot to a WhatsApp Cloud API instance fails with:

```
TypeError: t.mimetype?.startsWith is not a function
    at Ls.audioWhatsapp
    at ut.sendAttachment
    at ut.receiveWebhook
```

Text messages, images and PDFs are sent successfully. Only audio attachments are affected.

## Reproduction

1. Configure Chatwoot with Evolution API.
2. Configure a WhatsApp Business Cloud API instance.
3. Open a conversation.
4. Send an audio message from the Chatwoot interface.
5. The request fails before reaching Meta.

## Comparison

**Scenario A — audio sent directly using the Evolution API endpoint**
Result: ✅ No `mimetype?.startsWith` exception. Evolution continues processing normally.

**Scenario B — the exact same audio sent from Chatwoot**
Result: ❌ Exception thrown immediately, before the request reaches Meta:

```
TypeError: t.mimetype?.startsWith is not a function
```

## Payload comparison (anonymized)

**Works — direct Evolution API call:**

```json
{
  "mimetype": "audio/ogg"
  // ...
}
```

**Fails — via Chatwoot:**

```json
{
  "mimetype": null
  // or {}, or the actual captured value — TO FILL IN
}
```

*(placeholder — replace with the real captured object once logged; see note below)*

## Expected behavior

Evolution should correctly detect or normalize the attachment MIME type received from Chatwoot before calling `audioWhatsapp()`. Since `?.` only guards against `null`/`undefined`, a non-null, non-string `mimetype` (object, array, etc.) reaches `.startsWith()` unguarded and throws.

## Additional information

This appears to affect only the Chatwoot → Cloud API attachment flow. The direct Evolution API → Cloud API flow works correctly.

If necessary, I can provide the complete stack trace, the Chatwoot payload, the Evolution payload, and a full comparison between both requests.

### 🔄 Steps to Reproduce

1. Configure Chatwoot with Evolution API.
2. Configure a WhatsApp Business Cloud API instance.
3. Open a conversation in Chatwoot.
4. Send an audio message from the Chatwoot interface.
5. Request fails with TypeError before reaching Meta:
   TypeError: t.mimetype?.startsWith is not a function
       at Ls.audioWhatsapp
       at ut.sendAttachment
       at ut.receiveWebhook

Note: sending the same audio directly via the Evolution API endpoint (bypassing Chatwoot) works fine — no exception.

### ✅ Expected Behavior

Evolution should correctly detect or normalize the attachment MIME type received from Chatwoot before calling audioWhatsapp()

### ❌ Actual Behavior

TypeError: t.mimetype?.startsWith is not a function
    at Ls.audioWhatsapp
    at ut.sendAttachment
    at ut.receiveWebhook

### 🌍 Environment

- OS: [e.g. Ubuntu 20.04, Windows 10, macOS 12.0]
- Node.js version: [e.g. 18.17.0]
- Evolution API version: [e.g. 2.3.7]
- Database: [e.g. PostgreSQL 14, MySQL 8.0]
- Connection type: [e.g. Baileys, WhatsApp Business API]

### 📋 Logs

```shell

```

### 📝 Additional Context

_No response_

MatheusMartinho:
## 📋 Description

`mimeTypes.lookup()` returns `false`, not `undefined`, when it cannot resolve a type. Chatwoot serves attachments through Rails ActiveStorage URLs that carry no file extension, so for every Chatwoot audio the lookup in `processAudio()` returns `false` and `prepareMedia.mimetype` is set to `false`.

Line 1128 of `whatsapp.business.service.ts` then runs:

```ts
const isImage = message['mimetype']?.startsWith('image/');
```

Optional chaining only short-circuits on `null` and `undefined`. `false` is neither, so the call reaches `false.startsWith` and throws `TypeError: mimetype?.startsWith is not a function` before the request ever leaves for Meta.

This changes the check to an explicit type guard:

```ts
const isImage = typeof message['mimetype'] === 'string' && message['mimetype'].startsWith('image/');
```

## 🔗 Related Issue

Closes #2678

## 🧪 Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [ ] ⚡ Performance improvement
- [ ] 🧹 Code cleanup
- [ ] 🔒 Security fix

## 🧪 Testing

- [ ] Tested with different connection types (if applicable)

I reproduced the failure without a Chatwoot instance, by running the repository's own `mime-types` dependency against the exact expression from line 1128:

```
STEP 1 - what mimeTypes.lookup() returns
   false         Chatwoot (ActiveStorage, no extension)
   false         Chatwoot (bare blob)
   "audio/ogg"   normal URL with extension

STEP 2 - the object processAudio() returns
   {"mediaType":"audio","type":"link","mimetype":false}
   typeof message.mimetype: boolean

STEP 3 - line 1128 on develop, before the fix
   THROWS -> TypeError: message.mimetype?.startsWith is not a function
```

With the guard in place all four cases behave, and the existing behaviour for real MIME strings is unchanged:

| `mimetype` | `isImage` |
|---|---|
| `false` (Chatwoot) | `false` |
| `undefined` | `false` |
| `"audio/ogg"` | `false` |
| `"image/png"` | `true` |

Gates run locally on this branch (based on the current `develop` tip):

- `npm run lint:check` — clean
- `npm run build` (`tsc --noEmit` + `tsup`) — passes

I did not run `npm test`: the script points at `./test/all.test.ts`, which does not exist on `develop`.

## 📸 Screenshots (if applicable)

N/A — backend-only change.

## ✅ Checklist

- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published

## 📝 Additional Notes

**Why only audio is affected.** In `chatwoot.service.ts`, `sendAttachment()` downloads the file specifically to read its `Content-Type` when the extension-based lookup fails:

```ts
let mimeType = mimeTypes.lookup(parsedMedia?.ext) || '';

if (!mimeType) {
  const response = await axios.get(media, { responseType: 'arraybuffer' });
  mimeType = response.headers['content-type'] as string;
}
```

It then uses that value to route the message, but on the audio branch it passes only the raw URL to `audioWhatsapp()` and the resolved MIME type is dropped. `processAudio()` re-derives it from the same extension-less URL and gets `false`.

**This looks like an oversight rather than intent.** The line above, in `chatwoot.service.ts`, already normalizes the same call with `|| ''`. The `false` return value is understood elsewhere in the codebase, it is just not handled at this consumer. `processAudio()` and `processMedia()` even declare `let mimetype: string | false`, so the type is documented in the signature.

**Why the guard is here and not in the producers.** `startsWith('image/')` has exactly one call site in `src/`, this one. `processMedia()` (lines 1379 and 1383 on `develop`) uses the same `string | false` pattern and feeds the same line, so an image or document sent from an extension-less URL would hit the identical crash. One guard at the point of use covers every producer. Normalizing `false` to `undefined` in `processAudio()` alone would fix audio only.

**No behaviour change on the audio path.** For audio, `isImage` is only read inside a spread that already excludes `audio`, so the payload sent to Meta is byte-for-byte the same. The change removes the exception and nothing else.

Happy to move the normalization into `processAudio()` and `processMedia()` instead if you would prefer the producers to hand back a clean value, and equally happy to be told I have misread something. I am new to this codebase.

### DIFF
--- a/src/api/integrations/channel/meta/whatsapp.business.service.ts
+++ b/src/api/integrations/channel/meta/whatsapp.business.service.ts
@@ -1125,7 +1125,10 @@ export class BusinessStartupService extends ChannelStartupService {
           return await this.post(content, 'messages');
         }
         if (message['media']) {
-          const isImage = message['mimetype']?.startsWith('image/');
+          // `mimeTypes.lookup()` returns `false`, not `undefined`, when it cannot resolve the
+          // type, for example a Chatwoot ActiveStorage URL with no file extension. Optional
+          // chaining only guards `null` and `undefined`, so `false.startsWith` would throw.
+          const isImage = typeof message['mimetype'] === 'string' && message['mimetype'].startsWith('image/');
 
           content = {
             messaging_product: 'whatsapp',
--- a/src/api/integrations/channel/meta/whatsapp.business.service.ts
+++ b/src/api/integrations/channel/meta/whatsapp.business.service.ts
@@ -1125,9 +1125,8 @@ export class BusinessStartupService extends ChannelStartupService {
           return await this.post(content, 'messages');
         }
         if (message['media']) {
-          // `mimeTypes.lookup()` returns `false`, not `undefined`, when it cannot resolve the
-          // type, for example a Chatwoot ActiveStorage URL with no file extension. Optional
-          // chaining only guards `null` and `undefined`, so `false.startsWith` would throw.
+          // `mimeTypes.lookup()` returns `false`, not `undefined`, when the URL has no file
+          // extension, and optional chaining only guards `null` and `undefined`.
           const isImage = typeof message['mimetype'] === 'string' && message['mimetype'].startsWith('image/');
 
           content = {


MatheusMartinho:
Thanks for the review.

On the comment: agreed, trimmed it to the mechanism and dropped the Chatwoot-specific context, since this file is the Meta channel service.

On normalizing at assignment: I looked at whether the guards would be scattered, and `startsWith('image/')` has exactly one call site in `src/`, this one, so there is a single place to guard today.

I did consider normalizing at the producer. The reason I did not is that `processAudio()` and `processMedia()` both use the `string | false` pattern and both feed this same line, so fixing only `processAudio()` would leave `processMedia()` exposed, and fixing both means touching two producers to protect one consumer. I also leaned away from mapping `false` to `''` because an empty string asserts a known-empty type, while the real state is "unknown".

That said, this is a design call for the maintainers, not for me. If you would rather have the producers hand back a clean value, say the word and I will move the normalization into `processAudio()` and `processMedia()` and drop the guard here.

MatheusMartinho:
Thanks for the review — going through both points.

**On the comment**: already trimmed in 7336c3b3, it now just states the mechanism (`mimeTypes.lookup()` returns `false`, and optional chaining does not guard it) without the Chatwoot walkthrough.

**On normalizing at the assignment site**: I looked at doing that instead of guarding at the use site. `mimetype` is assigned from `mimeTypes.lookup()` in four places (`prepareMediaMessage` and `prepareAudioMessage`, each for the URL and filename branches) and then flows into `prepareMedia.mimetype`, which is also read as `contentType` when uploading media. Mapping `false` to `''` there is the deeper fix, but it changes what those other consumers receive, so it felt like a wider change than the crash this PR closes — hence the single guard at the one place that calls `.startsWith()`.

A helper felt like indirection for one call site, but if you'd rather have the normalization at the source I'm happy to push it in this PR — just say which shape you prefer.

MatheusMartinho:
Gentle ping on this one whenever someone from the team has a moment. It is a one-line guard for #2678: `mimeTypes.lookup()` returns `false` for Chatwoot's extensionless ActiveStorage URLs, and `false?.startsWith('image/')` still throws, so every Chatwoot audio sent through the Meta channel fails before the request goes out. The Sourcery points are addressed above, and all checks pass. Happy to rebase or change anything.

PR #2718, opened 2026-09-03
fix(chatwoot): handle missing remoteJidAlt for @lid contacts in createConversation

abdulsametkarakayali:
## Problem
When a WhatsApp contact sends messages with `addressingMode: 'lid'` but 
`remoteJidAlt` is missing/undefined from the message key, `createConversation` 
crashes with:

TypeError: Cannot read properties of undefined (reading 'split')

This causes 100% message loss (both text and reactions) for affected contacts 
in the Chatwoot integration — no conversation is ever created.

## Fix
- Fall back to `remoteJid` (the LID itself) when `remoteJidAlt` is absent, 
  instead of leaving `phoneNumber` as `undefined`.
- Made downstream `.split()` calls safe with optional chaining.
- Wrapped the LID contact-update block in try/catch so a failure there 
  logs a warning instead of aborting the whole conversation creation.

## Impact
Contacts without a resolvable `remoteJidAlt` will now show up in Chatwoot 
identified by their LID instead of a phone number, but messages will no 
longer be silently dropped.

Tested against v2.3.7.

## Summary by Sourcery

Handle incomplete WhatsApp LID contact identifiers so Chatwoot conversations continue to be created reliably.

Bug Fixes:
- Prevent Chatwoot conversation creation from failing when WhatsApp LID messages do not include a remote phone JID.
- Preserve message delivery by falling back to the LID identifier and handling missing JID components safely.
- Prevent failures in contact lookup and LID contact mapping from aborting conversation creation.

Enhancements:
- Avoid storing numeric LID identifiers as fake phone numbers in Chatwoot.
- Improve contact search compatibility and selection across Chatwoot response formats.

### DIFF
--- a/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
+++ b/src/api/integrations/chatbot/chatwoot/services/chatwoot.service.ts
@@ -632,8 +632,10 @@ export class ChatwootService {
   public async createConversation(instance: InstanceDto, body: any) {
     const isLid = body.key.addressingMode === 'lid';
     const isGroup = body.key.remoteJid.endsWith('@g.us');
-    const phoneNumber = isLid && !isGroup ? body.key.remoteJidAlt : body.key.remoteJid;
     const { remoteJid } = body.key;
+    // When addressingMode is lid, remoteJidAlt holds the phone JID.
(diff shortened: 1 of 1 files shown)


abdulsametkarakayali:
updt

Issue #2696, opened 2026-08-17
[BUG]  Pairing code is generated but WhatsApp fails to link the device on Evolution API v2.3.7


thiagovarallo:
### 📋 Bug Description

Hello,

I am running Evolution API v2.3.7 with the WHATSAPP-BAILEYS integration.

The pairing code is generated successfully, but the device cannot be linked to WhatsApp.

### 🔄 Steps to Reproduce

1. Create a new WhatsApp instance.
2. Request a pairing code using the phone number with country code.
3. Enter the generated 8-character pairing code in WhatsApp:
   Settings → Linked devices → Link a device → Link with phone number instead.
4. WhatsApp displays:
   "Couldn't link device. An error occurred. Please try again."

### ✅ Expected Behavior

After entering the generated pairing code, WhatsApp should complete the device linking and Evolution should emit a successful connection event.

### ❌ Actual Behavior

- Evolution generates a pairing code successfully.
- The webhook `qrcode.updated` is emitted.
- The pairing code and QR code are present in the webhook payload.
- WhatsApp rejects the device linking.
- The Evolution instance remains in `connecting`.
- No successful connection event is emitted.

### 🌍 Environment

- OS: [e.g. Ubuntu 20.04, Windows 10, macOS 12.0]
- Node.js version: [e.g. 18.17.0]
- Evolution API version: [e.g. 2.3.7]
- Database: [e.g. PostgreSQL 14, MySQL 8.0]
- Connection type: [e.g. Baileys, WhatsApp Business API]

### 📋 Logs

```shell
[ChannelStartupService]
instance: alpha
pairingCode: [REDACTED]
qrcodeCount: 6

[WebhookController]
event: qrcode.updated
instance: alpha
data.qrcode.pairingCode: [REDACTED]
data.qrcode.code: [REDACTED]
data.qrcode.base64: [PRESENT]
```

### 📝 Additional Context

- The code is entered immediately after generation.
- The phone number includes the country code and contains digits only.
- The same behavior happens with newly created instances.
- The instance remains stuck in `connecting`.
- Please let me know if this is a known compatibility issue with the current WhatsApp Web/Baileys protocol or if additional configuration is required.

qkoots:
Hello, I have the same issues.

diglesiasus:
I have the same issues too :(

favizz:
samee

Issue #2679, opened 2026-08-06
WhatsApp pairing always returns "Try again later" after scanning QR (Evolution API 2.3.2 / Baileys 6.7.21)


giovanealvesdemedeiros-cloud:
## Environment

- Evolution API 2.3.2
- Baileys 6.7.21
- Fresh installation
- Docker
- PostgreSQL
- Redis
- Ubuntu VPS (Hostinger)

## Problem

The instance is created successfully.

The QR Code is generated correctly.

The phone scans the QR Code successfully.

Immediately after scanning, WhatsApp shows:

> Try again later

The instance never reaches OPEN state.

## Logs

```
Error: Timed Out
error in sending keep alive
```

## Tests already performed

- Fresh PostgreSQL
- Fresh Redis
- Fresh Evolution installation
- Fresh database
- Fresh Docker image
- Baileys downgraded from rc.9 to 6.7.21
- HTTPS connectivity OK
- TLS connectivity OK
- DNS OK
- IPv4 OK
- IPv6 OK

The exact same behavior occurs with both:

- Evolution 2.3.7 (Baileys rc.9)
- Evolution 2.3.2 (Baileys 6.7.21)

Could someone help identify why WhatsApp always refuses the pairing with "Try again later"?

Additional information:

This problem has persisted for several weeks.

The QR Code is valid and is read successfully by WhatsApp.

The failure occurs only after the QR is accepted by the phone.

This suggests the failure happens during the authentication/session establishment step, not during QR generation.

PR #2727, opened 2026-09-10
fix(deps): patch baileys to restore WhatsApp device pairing (companion_reg_refresh)

cloviscoli:
## Problem

**Linking a new WhatsApp device is currently impossible** on Evolution — QR pairing never completes and the phone shows *"couldn't link device / não foi possível conectar no momento"*.

The root cause is upstream and **not an Evolution bug**: around 2026-07-28 WhatsApp added a new stage to the companion-registration flow, sending `<notification type='companion_reg_refresh'>` after the QR scan. Baileys acks it and discards it, so `pair-success` is never emitted and the QR ref pool drains. See [WhiskeySockets/Baileys#2737](https://github.com/WhiskeySockets/Baileys/issues/2737) — reproduced independently in **whatsmeow**, which confirms it affects every open client.

Because `companion_reg_refresh` is absent from Baileys rc13, rc14 *and* master, **no published Baileys release fixes this**. The fix only exists as unmerged PRs.

This is the umbrella issue for those reports (#2696, #2679 and others here).

## What this PR does

Uses the `patches/` mechanism that already exists on `develop` (`postinstall: patch-package`, currently holding only a README) to vendor the fix until upstream merges it.

1. **`package.json` / `package-lock.json`** — bumps `baileys` from `7.0.0-rc13` to `7.0.0-rc14` (latest published).
2. **`patches/baileys+7.0.0-rc14.patch`** (new) — applies two upstream fixes:
   - [Baileys#2765](https://github.com/WhiskeySockets/Baileys/pull/2765) — handles `companion_reg_refresh` by rotating the adv secret and re-rendering the QR without consuming a ref. **This is what restores pairing.**
   - [Baileys#2602](https://github.com/WhiskeySockets/Baileys/pull/2602) — guards `link_code_companion_reg`: WhatsApp sends it in two shapes, and the one without the crypto fields crashes with `Boom('Invalid buffer', 400)` (that's [#2600](https://github.com/WhiskeySockets/Baileys/issues/2600), which also breaks the pairing-code flow).

The patch touches only 4 files (`Socket/messages-recv.js`, `Socket/socket.js`, `Utils/companion-reg-client-utils.{js,d.ts}`) — no source maps, no unrelated build noise — and applies cleanly against a vanilla `baileys@7.0.0-rc14`.

## Verification

Tested on a real deployment (Evolution v2.3.7 base + this patch, Docker):

- **Before:** QR cycled every ~3.5 min, phone showed "couldn't connect", `connection.update` never reached `open`; `requestPairingCode()` crashed with `Invalid buffer`.
- **After:** paired on the second QR attempt, `state=open`, session stable, and a real inbound WhatsApp message was delivered through the webhook end-to-end.

I also posted the standalone Docker recipe in the linked comment for anyone who needs it before this lands.

## Notes for maintainers

- This vendors **unmerged** upstream PRs, so it should be dropped as soon as Baileys releases the fix — the patch filename is version-pinned, so a future baileys bump will surface it naturally.
- Happy to retarget this at `7.0.0-rc13` instead if you'd rather not bump the dependency in the same PR; the patch just needs regenerating against that version.

## Summary by Sourcery

Restore WhatsApp device and pairing-code support by updating Baileys and applying the required upstream fixes.

Bug Fixes:
- Restore WhatsApp device pairing by handling companion registration refresh notifications and preserving QR ref rotation.
- Prevent pairing-code flows from failing when companion registration messages lack expected cryptographic fields.

Enhancements:
- Vendor the required upstream Baileys fixes through a version-pinned patch until they are released upstream.

Build:
- Upgrade the Baileys dependency from 7.0.0-rc13 to 7.0.0-rc14.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "amqplib": "^2.0.1",
         "audio-decode": "^2.2.3",
         "axios": "^1.17.0",
-        "baileys": "7.0.0-rc13",
+        "baileys": "7.0.0-rc14",
         "class-validator": "^0.15.1",
         "compression": "^1.7.5",
         "cors": "^2.8.6",
@@ -5310,9 +5310,9 @@
       "license": "MIT"
     },
     "node_modules/baileys": {
-      "version": "7.0.0-rc13",
-      "resolved": "https://registry.npmjs.org/baileys/-/baileys-7.0.0-rc13.tgz",
-      "integrity": "sha512-v8k74K8B5R7WNYGa26MyJAYEu3Wc4BSuK01QaK8lr30lhE8Nga31nWNu8KN0NDDt+Fsvkq4SQFFI8Q13ghjKmA==",
+      "version": "7.0.0-rc14",
+      "resolved": "https://registry.npmjs.org/baileys/-/baileys-7.0.0-rc14.tgz",
+      "integrity": "sha512-pewtrljhWx5JTUBvvkXZz1fL3JPiwzjBsnhx/DWf2LWBx1cZNH7J/sF22xDehba5mySWUQU4a1Pwt7lWARUxaA==",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "amqplib": "^2.0.1",
     "audio-decode": "^2.2.3",
     "axios": "^1.17.0",
-    "baileys": "7.0.0-rc13",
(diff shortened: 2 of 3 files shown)


otaviovicario:
+1 on getting this merged — hitting the exact issue this PR addresses. Self-hosted v2.3.7, new device pairing via QR fails with WhatsApp's "can't connect new devices" error since late July, while existing sessions keep working fine. This is blocking new customer onboarding for us.

I can build this branch and test it against a real instance if that helps move it along — will report back with results either way.

PR #2731, opened 2026-09-21
fix(chatwoot): resolve instanceId before looking up reaction targets

ticczaleski:
## Summary
- `ChatwootRouter`'s webhook route builds `instance` from the URL's `:instanceName` param alone (`RouterBroker#dataValidate` never sets `instanceId`). The `message_created` path patches `instance.instanceId = waInstance.instanceId` right after resolving the running instance, but `handleReactionWebhook` never did — so `getMessageByKeyId`'s `WHERE "instanceId" = ${instance.instanceId}` predicate always compared against `undefined` and matched nothing.
- Effect in production: every agent reaction sent from the Chatwoot dashboard logged `Could not resolve WhatsApp key for chatwoot message X; acknowledging without retry` and was silently dropped, even though the target message existed in Evolution's local `Message` table. Confirmed live: the reaction webhook fired correctly from Chatwoot (capability gate, event dispatch, delivery all healthy), reached Evolution, but failed the lookup — while the reverse direction (WhatsApp contact reaction -> Chatwoot) already worked since it runs through the in-process `messages.upsert` handler, which does set `instanceId` correctly.
- Fix: fill in `instance.instanceId` from `waInstance.instanceId` in `handleReactionWebhook`, same as the sibling code path.

## Test plan
- [x] `npx vitest run src/api/integrations/chatbot/chatwoot/services/chatwoot-reactions.spec.ts` — 10/10 passing, including a new regression test that reproduces a bare `{instanceName}` instance (as the real webhook route produces) and asserts `getMessageByKeyId` is called with the resolved `instanceId`.
- [x] `eslint` / `tsc --noEmit` via pre-commit hook — clean.
- [ ] Manual: verified on production (`evolution-api-ti1`) that before this fix, an agent reaction on a real message never reached WhatsApp; the code path was confirmed via live logs against message id 1246 which existed in `Message` with the correct `key.id` but was never found due to the `instanceId` mismatch.

## Summary by Sourcery

Improve Chatwoot delivery reliability and reaction synchronization across Chatwoot and WhatsApp.

New Features:
- Support bidirectional Chatwoot and WhatsApp reactions, including reaction creation, updates, and removals.
- Preserve WhatsApp message identifiers on Chatwoot messages to support quoted-message resolution.

Bug Fixes:
- Resolve the running WhatsApp instance ID before looking up Chatwoot reaction targets, preventing valid targets from being missed.
- Report outbound delivery failures on the original Chatwoot message instead of creating separate private notes.
- Prevent duplicate Chatwoot webhook deliveries from sending duplicate WhatsApp messages or reactions.

Enhancements:
- Add cache-backed atomic delivery claims with local and Redis implementations.
- Improve quoted-message lookup by preferring WhatsApp-native external identifiers.

Build:
- Switch the test scripts to Vitest and add Vitest configuration and dependencies.

CI:
- Update Docker publishing workflows with the new image namespace, manual tagging, version and SHA tags, concurrency controls, and build caching.

Tests:
- Add regression and service tests covering reaction bridging, delivery handling, cache claims, and instance ID resolution.

Chores:
- Update package lockfile dependencies.


ticczaleski:
Opened by mistake against the wrong repo (targeted the upstream project instead of our internal fork). Closing; the real PR is in our fork.

PR #2509, opened 2026-04-20
fix(baileys): prevent healthy instances from being killed after stream:error 515

octo-patch:
Fixes #2498

## Problem

When WhatsApp sends `stream:error code=515` (Connection Replaced — normal multi-device protocol behavior), Baileys handles the reconnect correctly and fires `connection.update` with `state='open'`. However, WhatsApp then sends a `401 loggedOut` message to clean up the old session slot. Evolution API's close handler incorrectly treated this 401 as a real logout, killing the newly-connected healthy instance.

Sequence of events that triggered the bug:
1. `stream:error code=515` — WhatsApp notifies of connection replacement
2. Baileys reconnects automatically (~5s)
3. `CONNECTED TO WHATSAPP ✓` — session is open, webhook fires `state='open'`
4. WhatsApp sends `401 loggedOut` to clean up old session slot
5. Evolution API sees 401 → triggers `logout.instance` → **REMOVED** ← bug

## Solution

Track when a `stream:error 515` node arrives via the `CB:stream:error` WebSocket event using a class-level timestamp (`_lastStream515At`). In the connection close handler, if a `loggedOut (401)` event fires within 30 seconds of a 515, treat it as a transient reconnect side-effect rather than a real logout and call `connectToWhatsapp()` instead of destroying the instance.

Changes:
- Added `private _lastStream515At = 0;` class property to track when 515 last occurred
- Added `CB:stream:error` WebSocket handler in `createClient()` to record the timestamp when code is `'515'`
- Updated `shouldReconnect` logic in the connection close handler to allow reconnect when a loggedOut follows within 30s of a 515

## Testing

The fix can be validated by:
1. Connecting a WhatsApp instance via QR code
2. Observing Docker logs: if `stream:error code=515` appears followed by `CONNECTED TO WHATSAPP ✓`, the instance should remain connected instead of being removed
3. Confirming no `WAMonitoringService: REMOVED`/`LOGOUT` appears after a successful reconnect

## Summary by Sourcery

Bug Fixes:
- Prevent instances from being destroyed when a loggedOut (401) event occurs shortly after a WhatsApp stream:error 515 reconnect sequence.

PR #2501, opened 2026-04-08
fix: logout instance

daniloleonecarneiro:
## 📋 Description

Improves the reliability of the WhatsApp instance logout flow and prevents crashes during connection lifecycle events.

**Changes in `BaileysStartupService`:**

- `logoutInstance()`: sets `endSession = true` before teardown so reconnection is suppressed; wraps `client.logout()` and `client.ws.close()` in individual try/catch blocks to gracefully handle already-disconnected states; forces `stateConnection` to `{ state: 'close', statusReason: 401 }` and updates the instance `connectionStatus` to `'close'` in the database after cleanup.
- Adds status code `408` to `codesToNotReconnect` to avoid reconnection loops on request-timeout disconnects.
- Adds a null guard for `client.user?.id` on `connection === 'open'` to prevent a `TypeError` crash when Baileys fires the open event before the user object is populated.

## 🔗 Related Issue

Closes #(issue_number)

## 🧪 Type of Change
- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [ ] ⚡ Performance improvement
- [ ] 🧹 Code cleanup
- [ ] 🔒 Security fix

## 🧪 Testing
- [x] Manual testing completed
- [x] Functionality verified in development environment
- [x] No breaking changes introduced
- [x] Tested with different connection types (if applicable)

## 📸 Screenshots (if applicable)

N/A

## ✅ Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have manually tested my changes thoroughly
- [x] I have verified the changes work with different scenarios
- [x] Any dependent changes have been merged and published

## 📝 Additional Notes

The `client.logout()` call can throw when the WebSocket is already closed (e.g. after a network drop). Previously this would leave the instance in an inconsistent state in the database. The new flow ensures the database record always reflects `connectionStatus: 'close'` regardless of whether the Baileys client was reachable during logout.

## Summary by Sourcery

Improve reliability and safety of WhatsApp instance logout and connection handling.

Bug Fixes:
[...]

PR #2732, opened 2026-09-21
fix: reconnect on 408 with bounded backoff instead of destroying the session

maonamassaaceleradora:
## 📋 Description

Baileys reports `408` (`DisconnectReason.connectionLost` / `timedOut`) when the 30s keep-alive detects a socket that stopped answering. That is a transient network condition, not a real logout.

`408` was added to `codesToNotReconnect` in #2501 to avoid reconnect loops on transient drops. The side effect is that a timed-out keep-alive now takes the same path as a real logout: `logout.instance` fires, `cleaningUp()` runs, and the persisted session rows and instance directory are deleted. The credentials are destroyed and the only way back is a fresh QR scan.

**Production evidence:** one of our production instances closed with `428`, reconnected on its own and was back to `open` seven seconds later — normal transient-drop behaviour. Twenty minutes later a `408` close on the same instance went straight to `LOGOUT` and wiped its credentials, although the account itself was still perfectly valid.

This keeps the intent of #2501 (bound the reconnect loop) without paying for it with destroyed credentials on every timed-out keep-alive.

## What changes

- `408` leaves `codesToNotReconnect`. Terminal codes are now `loggedOut`, `forbidden`, `402` and `406` — codes where the credentials really are invalid.
- Transient closes retry with an exponential backoff ladder (`3s → 6s → 12s → 30s → 60s → 120s → 300s`), capped at 20 attempts. That hard attempt limit is what actually bounds the loop #2501 was guarding against, rather than treating one status code as unrecoverable.
- When the ladder is exhausted the instance is left `closed` **without** emitting `logout.instance`, so the persisted session survives and `/instance/connect` is enough to bring it back — no QR re-scan.
- The attempt counter resets on a healthy `open` and on any externally triggered connect (`/instance/connect`, boot auto-connect), but not on the ladder's own scheduled retries — otherwise the ladder would never exhaust.

## 🔗 Related Issue

No existing issue for this specific regression. Related prior work: #2501 (introduced `408` as terminal) and #2509 (added the `stream:error 515` reconnect grace window this change sits alongside).

## 🧪 Type of Change

## 🧪 Testing

- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] Manual testing: reproduced the production timeline (transient `428` reconnect, followed by `408`) and confirmed the instance now recovers through the backoff instead of being logged out.
- [ ] Automated coverage for the close handler — none existed before this change either. Happy to add it if maintainers want that as a merge condition.

## ✅ Checklist

- [x] I have verified the changes work with different scenarios (`428` → healthy reconnect; `408` → backoff → recovered; ladder exhaustion → closed with credentials preserved)

## 📝 Additional Notes

This targets `develop` (2.4.0), not `main` (2.3.7). `main` predates both #2501 and #2509 and does not have `408` in `codesToNotReconnect` at all — but it also has none of the backoff, attempt-cap or 515-grace machinery this fix builds on, so the change is not portable to `main` as-is.

## Summary by Sourcery

Restore automatic recovery from transient WhatsApp disconnects while bounding retries and protecting valid session credentials.

Bug Fixes:
- Preserve persisted WhatsApp sessions when transient 408 connection timeouts occur instead of treating them as logouts that require QR re-pairing.

Enhancements:
- Add bounded exponential backoff for transient reconnects, with attempt limits and reset behavior after healthy or externally initiated connections.
- Leave instances closed but recoverable with credentials intact when transient reconnect attempts are exhausted, while retaining cleanup for genuinely terminal authentication failures.


maonamassaaceleradora:
Good catch — addressed in 4fc76609.

The retry timer was never cancelled, and the backoff ladder made that window much wider than the previous fixed 3s delay: a logout during a 5-minute wait would still see the timer fire and rebuild the socket after credentials were gone.

`connectionUpdate` already refuses to act when `isDeleting`/`endSession` is set, so the fix applies that same rule at the moment the timer fires, and `logoutInstance()` now clears the pending handle and resets the ladder state.

